PR #1, opened 2025-05-13
All blocks now output their isvalid signal in their return type

jscatena88:
I added the signal in a few blocks and made an f64 "bool" output into an actual bool. I'm not 100% sure that will work for this next phase of codegen changes but it shouldn't be too bad to update them all to output an f64 instead of a bool down the line if need be now that they all have the bool at least


Dillonmcewan (approved):
LGTM

PR #5, opened 2025-05-16
Custom Block traits Interop with Pass Types

jscatena88:
Some trait implementations to interop custom blocks with the new signal data types.


Dillonmcewan (approved):
LGTM!

PR #6, opened 2025-05-29
Changes to Support Codegen Changes

jscatena88:
These changes primarily make the input types of blocks more flexible. Blocks that previously only accepted `bool` input can now take in any Scalar value. In those cases `0` is False and all other values are true. 

We also return a float in lieu of a `bool` in a few places, this is to support codegen changes in the main application, but will not be permanent.

### DIFF
--- a/block-data/src/lib.rs
+++ b/block-data/src/lib.rs
@@ -38,6 +38,16 @@ pub struct BlockData {
     _type: BlockDataType,
 }
 
+impl BlockDataRead for BlockData {
+    fn get_scalar(&self) -> f64 {
+        self.scalar()
+    }
+
+    fn get_matrix(&self) -> (usize, usize, &[f64]) {
+        (self.nrows(), self.ncols(), self._data.as_slice())
+    }
+}
+
 impl BlockDataRead for &BlockData {
     fn get_scalar(&self) -> f64 {
         self.scalar()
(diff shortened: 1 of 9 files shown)


block-data/src/lib.rs
```diff
@@ -38,6 +38,16 @@ pub struct BlockData {
     _type: BlockDataType,
 }
 
+impl BlockDataRead for BlockData {
```
Dillonmcewan, line 41: Do you know of a good pattern to do this so you don't have to reimpl a trait for every combination of ref/non-ref? Especially annoying when you have a generic in the trait and you end up potentially needing:
```
impl Foo<Bar> for Baz {}
impl Foo<&Bar> for Baz {}
impl Foo<Bar> for &Baz {}
impl Foo<&Bar> for &Baz {}
```
No need to change here since it's only one extra copy and BlockData is not long for this world, but something I've run into in a few places.

Dillonmcewan (approved):
LGTM with one tangential question

block-data/src/lib.rs
jscatena88: Unfortunately I don't know a great way to do that, I get why they are separate  since having a trait available on both isn't always ideal anyway but it would be nice to have some way to indicate that you just want it for both.

PR #15, opened 2025-05-30
Multi-channel PWM for STM32

amcelroy:
Adds support for multiple channel PWM for STM32 timers.

Screenshot 2025-05-30 at 10 27 02 AM

Screenshot 2025-05-30 at 12 15 03 PM


pictorus-stm32/src/pwm_protocol.rs
```diff
@@ -54,57 +53,196 @@ impl<T: timer::GeneralInstance4Channel> Pwm for PwmWrapper<'_, T> {
...
+        self.set_duty_ch1(duty_cycle.0);
+        self.set_duty_ch2(duty_cycle.1);
+        self.set_duty_ch3(duty_cycle.2);
+        self.set_duty_ch4(duty_cycle.3);
+    }
+
+    fn enable_ch1(&mut self) {
+        if self.ch1.is_some() {
+            self.simple_pwm.enable(self.ch1.unwrap());
+        }
```
Dillonmcewan, line 119: Probably cleaner to do something like:
```
if let Some(ch) = self.ch1 {
  self.simple_pwm.enable(ch);
}
```

That way you don't need an unwrap

Dillonmcewan: Also wondering if rather than having explicit `<thing>_ch1`, `<thing>_ch2` methods it would be better to have a single:
```rust
fn enable_channel(&mut self, &channel: &Option<Channel>) {
    if let Some(ch) = channel {
        self.simple_pwm.enable(ch);
   }
}
```

Which you could call as:
```rust 
self.enable_channel(&self.ch1);
self.enable_channel(&self.ch2);
// etc.
```

That way you have less boilerplate/redundant methods

pictorus-stm32/src/pwm_protocol.rs
```diff
@@ -114,16 +252,28 @@ impl<T: timer::GeneralInstance4Channel> OutputBlock for PwmWrapper<'_, T> {
...
 
         if (self.get_period() - period).abs() >= PWM_PERIOD_TOLERANCE_POINT_1_US {
             self.set_period(period);
         }
 
-        if (self.get_duty(()) - duty_cycle).abs() >= PWM_DUTY_CYCLE_TOLERANCE_16_BIT {
-            self.set_duty((), duty_cycle);
+        if (self.get_duty_ch1() - duty_cycle1).abs() >= PWM_DUTY_CYCLE_TOLERANCE_16_BIT {
+            self.set_duty_ch1(duty_cycle1);
+        }
```
Dillonmcewan, line 265: Nit: could wrap these into a method:
```rust
fn maybe_update_duty_cycle(&mut self, &channel: Option<Channel>, duty: f64) {
    if (self.get_duty(channel) - duty).abs() >= PWM_DUTY_CYCLE_TOLERANCE_16_BIT {
        self.set_duty(channel, duty);
    }
}
```

and then:
```rust
self.maybe_set_duty(&self.ch1, duty_cycle1);
self.maybe_set_duty(&self.ch2, duty_cycle2);
```

Dillonmcewan (commented):
Functionality looks good, but I think a lot of room to consolidate these methods

pictorus-stm32/src/pwm_protocol.rs
amcelroy: Rust doesn't like a function that takes a mut self passing in a reference to self as an argument, like for the channels. I went with channel copies instead of references for the channels to avoid this issue and reduce the verbosity.

Dillonmcewan: Ah, yeah looks like that `Channel` object is just an enum, so passing by value seems like the better option anyway

Dillonmcewan (approved):
Looks good!

PR #17, opened 2025-06-20
Replace JSON with Postcard bytestreams for telemetry encoding and transmission in pictorus-internal

amcelroy:
Alters telemetry encoding to use the [Postcard format](https://docs.rs/postcard/latest/postcard/) instead of JSON to reduce extraneous data, like key names, and utilize byte encoding instead of string representation of values.

- Removes miniserde and replaces with serde, serde_json (still needed to generate CSV files in Linux), and Postcard
- Adds a new Encoder trait that should encode a Serializable object and returns a heapless::Vec<u8, N>
- Removes `PictorusLogger`, use the `Logger` trait instead of both `PictorusLogger` and `Logger`
- RttLogger now creates 2 up channels for use over RTT, a data channel (index 0, defaults size 1024 bytes), and a log channel (index 1, defaults to 512 bytes). The Data channel is used for writing raw `&[u8]`, the log channel is used with `rprint` and `rprintln`. 
- RttLogger and UdpLogger now emit Postcard Wire Format byte streams. These can be decoded using the Rust struct in the generated code, or using the rules defined in the [Postcard Wire Format](https://postcard.jamesmunns.com/wire-format).

### DIFF
--- a/pictorus-internal/Cargo.toml
+++ b/pictorus-internal/Cargo.toml
@@ -37,5 +37,5 @@ heapless = { version = "0.7.0" }
 temp-env = "0.3"
 
 [features]
-std = ["dep:serde/std", "dep:env_logger", "dep:chrono", "dep:serde_json", "dep:postcard"]
+std = ["serde/std", "dep:env_logger", "dep:chrono", "dep:serde_json", "dep:postcard"]
 rtt = ["dep:rtt-target", "dep:postcard", "dep:serde"]
--- a/pictorus-internal/Cargo.toml
+++ b/pictorus-internal/Cargo.toml
@@ -31,12 +31,10 @@ rtt-target = { git = "https://github.com/Pictorus-Labs/rtt-target", branch = "al
 postcard = { version = "1.1.1", features = ["alloc"] }
 serde = { version = "1.0.*", default-features = false, features = ["derive", "alloc"] }
 serde_json = { version = "1.0", default-features = false, features = ["alloc", "preserve_order"] }
-heapless = "0.8.0"
 
 [dev-dependencies]
 temp-env = "0.3"
 
 [features]
-default = ["rtt"]
 std = ["serde/std", "dep:env_logger", "dep:chrono"]
 rtt = ["dep:rtt-target"]
--- a/pictorus-internal/Cargo.toml
+++ b/pictorus-internal/Cargo.toml
@@ -28,13 +28,13 @@ nb = "1.1.0"
 chrono = { version = "0.4.40", optional = true }
 env_logger = { version = "0.11.8", optional = true }
 rtt-target = { git = "https://github.com/Pictorus-Labs/rtt-target", branch = "alignment-fix", optional = true }
-postcard = { version = "1.1.1", features = ["alloc"] }
-serde = { version = "1.0.*", default-features = false, features = ["derive", "alloc"] }
+postcard = { version = "1.1.1", features = ["alloc"], optional = true }
+serde = { version = "1.0.*", default-features = false, features = ["derive", "alloc"], optional = true }
 serde_json = { version = "1.0", default-features = false, features = ["alloc", "preserve_order"], optional = true }
 
 [dev-dependencies]
 temp-env = "0.3"
 
 [features]
-std = ["serde/std", "dep:env_logger", "dep:chrono", "dep:serde_json"]
-rtt = ["dep:rtt-target"]
+std = ["serde/std", "dep:env_logger", "dep:chrono", "dep:serde_json", "dep:postcard"]
+rtt = ["dep:rtt-target", "dep:postcard", "dep:serde"]
--- a/pictorus-internal/Cargo.toml
+++ b/pictorus-internal/Cargo.toml
@@ -28,9 +28,10 @@ nb = "1.1.0"
 chrono = { version = "0.4.40", optional = true }
 env_logger = { version = "0.11.8", optional = true }
 rtt-target = { git = "https://github.com/Pictorus-Labs/rtt-target", branch = "alignment-fix", optional = true }
-postcard = { version = "1.1.1", features = ["alloc"], optional = true }
+postcard = { version = "1.1.1", optional = true }
 serde = { version = "1.0.*", default-features = false, features = ["derive", "alloc"], optional = true }
 serde_json = { version = "1.0", default-features = false, features = ["alloc", "preserve_order"], optional = true }
+heapless = { version = "0.7.0" }
 
 [dev-dependencies]
 temp-env = "0.3"
--- a/pictorus-internal/Cargo.toml
+++ b/pictorus-internal/Cargo.toml
@@ -37,5 +37,5 @@ heapless = { version = "0.7.0" }
 temp-env = "0.3"
 
 [features]
-std = ["serde/std", "dep:env_logger", "dep:chrono", "dep:serde_json", "dep:postcard"]
+std = ["dep:serde/std", "dep:env_logger", "dep:chrono", "dep:serde_json", "dep:postcard"]
 rtt = ["dep:rtt-target", "dep:postcard", "dep:serde"]
--- a/pictorus-internal/Cargo.toml
+++ b/pictorus-internal/Cargo.toml
@@ -35,6 +35,7 @@ heapless = { version = "0.7.0" }
 
 [dev-dependencies]
 temp-env = "0.3"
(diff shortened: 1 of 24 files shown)


pictorus-internal/src/loggers/csv_logger.rs
```diff
@@ -68,77 +55,76 @@ impl Logger for CsvLogger {
...
-                        // Simple scalar values don't need any special formatting
-                        sample.push_str(&miniserde::json::to_string(field_inner.1));
-                    }
+    let json = serde_json::to_value(data).unwrap();
+    if let Some(json_map) = json.as_object() {
+        for (_, value) in json_map {
+            match value {
+                serde_json::Value::Null => {}
+                serde_json::Value::Bool(_) => {
+                    // This indicates a boolean value, we just have it be empty
```
jscatena88, line 99: These comments are a bit wonky, they're not empty they are just being passed through unaltered right?

jscatena88: Also a bit odd that Bool calls to_string on `value` and Number calls it on the contained `number: Number` value. Is that required for serde_json?

pictorus-internal/src/loggers/csv_logger.rs
```diff
@@ -68,77 +55,76 @@ impl Logger for CsvLogger {
...
-        }
-        _ => {
-            panic!("Unsupported data format for CSV samples");
+            sample.push(',');
         }
     }
+
+    if sample.ends_with(',') {
+        sample.pop(); // Remove the trailing comma
+    }
```
jscatena88, line 126: same super nit on this being a bit of a code smell to me (although not messing with manually calling `next()` on an iterator is definitely an improvement

pictorus-internal/src/loggers/udp_logger.rs
```diff
@@ -15,6 +19,7 @@ pub struct UdpLogger {
     publish_socket: String,
     last_udp_publish_time: Option<Duration>,
     has_udp_connection: bool,
+    encoder: PostcardEncoderCOBS,
```
jscatena88, line 22: Totally not needed for this PR,  but the way you split the design up here with the PictorusEncoder trait made me think this would be generic over a `T: PictorusEncoder`.  If we don't think we'll be doing that in the near-ish term I'd be tempted to not define the `PictorusEncoder` trait since it could be a bit of a pre-optimization

amcelroy: Yeah, I'm ok with that

jscatena88:
Overall a great improvement and simplification. I haven't started the companion PRs but if I remember correctly Logger was being used for Error message strings and the like and PictorusLogger was being used for telemetry data specifically. With the new changes Logger seems to only be taking on the Telemetry data specific role? Is that the case? Again, I am assuming it will be more clear as I read the other PRs, but presumably we will be using raw `warn!`, `rprintln!`/`println!` or something similar to back fill that usecase

pictorus-internal/src/loggers/csv_logger.rs
amcelroy: Using the String(value) instead of serde_json::to_string is causing:
```
failures:

---- loggers::csv_logger::tests::test_csv_formatting_empty_fields stdout ----

thread 'loggers::csv_logger::tests::test_csv_formatting_empty_fields' panicked at pictorus-internal/src/loggers/csv_logger.rs:189:9:
assertion `left == right` failed
  left: "main_state,1.234,2.234,,,,"
 right: "\"main_state\",1.234,2.234,,,,"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- loggers::csv_logger::tests::test_csv_formatting stdout ----

thread 'loggers::csv_logger::tests::test_csv_formatting' panicked at pictorus-internal/src/loggers/csv_logger.rs:164:9:
assertion `left == right` failed
  left: "main_state,1.234,2.234,\"[[0.0,2.0,4.0]]\",1.0,\"[[5.0,6.0],[7.0,8.0]]\",\"[1,2,3]\""
 right: "\"main_state\",1.234,2.234,\"[[0.0,2.0,4.0]]\",1.0,\"[[5.0,6.0],[7.0,8.0]]\",\"[1,2,3]\""


failures:
    loggers::csv_logger::tests::test_csv_formatting
    loggers::csv_logger::tests::test_csv_formatting_empty_fields

test result: FAILED. 34 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

Looks like an issue with `\"`? I will convert them all to `&serde_json::to_string(value).unwrap()` for consistency.

pictorus-internal/Cargo.toml
```diff
@@ -22,17 +22,20 @@ num-traits = { version = "0.2.19", default-features = false, features = [
   "libm",
 ] }
 log = "0.4.21"
-miniserde = { version = "0.1.42", default-features = false }
```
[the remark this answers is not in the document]
Dillonmcewan: Why do we still need miniserde for the JSON block? Could we just switch that to serde as well?

pictorus-internal/src/loggers/rtt_logger.rs
```diff
@@ -1,26 +1,59 @@
...
     last_broadcast_time: Option<Duration>,
     previous_heap_used: usize,
     last_heap_log_time: Duration,
+    data_channel: UpChannel,
+    encoder: PostcardEncoderCOBS,
 }
 
 impl RttLogger {
     pub fn new(publish_period: Duration) -> RttLogger {
+        let data_channel = pictorus_rtt_init();
```
Dillonmcewan, line 49: Maybe @jscatena88 can weigh in here, but I generally like to have any resource allocation/peripheral configuration happen in the main.rs file of the app. It would be really easy for someone not to realize the RTT was getting set up here and stomp on it with some other configuration. This is the way I've seen it done with other tools that utilize RTT (other than defmt which is doing a lot of custom stuff)

pictorus-internal/src/loggers/udp_logger.rs
Dillonmcewan: Agreed. Think it makes sense to leave for now, but this sets us up nicely if we want to support different serialization formats

Dillonmcewan (commented):
Couple minor comments

pictorus-internal/src/loggers/rtt_logger.rs
amcelroy: Yes, I recall talking about it and tried a few things, but kept coming back to having to create `RttLogger` (or the RTT channels) in main and then pass either of those into the `AppInterface` where it would become part of the `AppInterface` struct and it didn't seem as clean.

If it is preferable, I can re-work the `AppInterface` codegen and have the `RttLogger` created at the top level and passed in if logging is enabled in the sibling PR.

Dillonmcewan: Ah yeah that's fair. If it's gonna add more complexity I think fine to leave as-is for now. It would be nice if there was a way for us to raise a compile error if someone tries to pull in rtt target upstream of this, but not something we need to deal with immediately

pictorus-internal/Cargo.toml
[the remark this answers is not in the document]
jscatena88: Crazy how serde seems to compile smaller than miniserde. I agree with Dillon that getting rid of miniserde entirely is definitely going to be the move. I'm personally ok for now just making that an issue on the backlog though. It'd be a nice self contained small-ish task for one of us to pickup if we have a random free day we need to fill

pictorus-internal/src/encoders/postcard_encoder.rs
```diff
@@ -0,0 +1,40 @@
...
+    #[test]
+    fn test_postcard_encoder() {
+        let log_data = LogData { app_time: 1.0 };
+
+        let mut encoder = PostcardEncoderCOBS {};
+        let encoded = encoder.encode::<64>(&log_data);
+
+        // COBS encoding removes zero bytes and adds a sentinel byte, so the length of an 8 byte float
+        // with some 0's in it will be 10 bytes.
+        assert!(encoded.len() == 10);
```
Dillonmcewan, line 38: Agreed with Jason that this test could be improved. It would be more useful to encode a few different data types and then ensure they deserialize as expected

Dillonmcewan (approved):
LGTM with one suggestion for improving tests

PR #18, opened 2025-07-01
Rust 1.88 Clippy updates

amcelroy:
Variables can be used in print macros now instead of comma separated after the macro string.


Dillonmcewan (approved):
LGTM

PR #21, opened 2025-07-22
Panic in Timing::new if a sub-microsecond timestep is requested

jscatena88:
If too small of a timestep is asked for the logic of some of our timing code begins to breakdown because the `Timing::timestep_us` member variable gets set to 0. While we could potentially move to support sub-microsecond timesteps in the future if requested for now Timing has been modified to panic in this situation. This prevents undefined behavior during runtime instead panicking at program startup if unsupported timing parameters are requested.


amcelroy (approved):
Thanks for checking on negative hertz, should we add as a unit test? Otherwise, LGTM.

Dillonmcewan (approved):
LGTM

PR #22, opened 2025-08-01
More efficient CSV logger

cwsullivan:
# Optimize CSV Logger Performance

## Motivation

The CSV logger was identified as a performance bottleneck in Monte Carlo sims. Profiling revealed that the logger was spending excessive time in JSON serialization and file I/O operations, significantly impacting overall application performance.

## Root Cause Analysis

The original implementation had several performance issues:

1. **Double JSON serialization**: Data was serialized to JSON twice - once to extract structure and again for each value
2. **Unbuffered file writes**: Each log entry resulted in a direct system call to write to disk
3. **String allocations**: New strings were allocated for every log entry
4. **Redundant serialization**: Using `serde_json::to_string()` for simple types like numbers and booleans

### 1. Added Buffered I/O
- Replaced direct `File` writes with `BufWriter` using a 64KB buffer
- Reduces system calls by ~100x (depending on log entry size)
- Buffer size chosen to match typical filesystem block sizes

### 2. Eliminated Double Serialization
- Introduced `write_csv_direct()` method that serializes to JSON only once for structure
- Formats values directly to CSV without intermediate JSON strings
- Special handling for numbers and arrays to avoid `to_string()` overhead

### 3. String Buffer Reuse
- Pre-allocated string buffer that's cleared and reused for each log entry
- Eliminates allocation overhead in the hot path

### 4. Automatic Flush on Drop
- Implemented `Drop` trait to ensure buffered data is written even if logger is dropped unexpectedly
- Prevents data loss in error scenarios

## Testing

Added comprehensive test coverage for:
- Actual file output verification (original tests only used `/dev/null`)
- Header deduplication
- Buffer reuse and data integrity
- Flush-on-drop behavior
- Large array handling performance
- Zero logging period edge case

All existing tests continue to pass.

### DIFF
--- a/pictorus-internal/src/loggers/csv_logger.rs
+++ b/pictorus-internal/src/loggers/csv_logger.rs
@@ -1,39 +1,37 @@
 use chrono::Utc;
 use core::time::Duration;
 use log::info;
-use std::io::Write;
+use std::io::{BufWriter, Write};
 use std::{fs::File, string::String};
+use alloc::vec::Vec;
 
 use super::Logger;
 
-/// CsvLogger logs data to a file in CSV format.
-///
-/// Note, this uses a UTC time to be passed into the log. Other loggers
-/// may use the app time in conjunction with the a device manager starting
-/// timestamp to calculate the UTC time.
 pub struct CsvLogger {
     last_csv_log_time: Option<Duration>,
     pub csv_log_period: Duration,
-    pub file: std::fs::File,
+    pub writer: BufWriter<File>,
     pub output_path: std::path::PathBuf,
     pub app_start_epoch: Duration,
+    header_written: bool,
+    buffer: String,
 }
 
 impl CsvLogger {
     pub fn new(csv_log_period: Duration, output_path: std::path::PathBuf) -> Self {
-        let mut file_obj = File::create("/dev/null").unwrap();
-        if !csv_log_period.is_zero() {
+        let file = if !csv_log_period.is_zero() {
             info!("DataLogger CSV output period: {csv_log_period:?}");
             info!("Streaming data output to file: {}", output_path.display());
-            file_obj = File::create(std::path::PathBuf::from(&output_path)).unwrap();
+            File::create(&output_path).unwrap()
         } else {
             info!("Not streaming output to file, logging rate set to zero.");
-        }
+            File::create("/dev/null").unwrap()
+        };
 
         CsvLogger {
             last_csv_log_time: None,
             csv_log_period,
-            file: file_obj,
+            writer: BufWriter::with_capacity(65536, file), // 64KB buffer
             output_path,
             app_start_epoch: Duration::from_micros(
                 Utc::now()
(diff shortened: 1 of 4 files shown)


pictorus-internal/src/loggers/csv_logger.rs
```diff
@@ -14,33 +14,123 @@ use super::Logger;
...
             info!("Not streaming output to file, logging rate set to zero.");
-        }
+            File::create("/dev/null").unwrap()
+        };
 
         CsvLogger {
             last_csv_log_time: None,
             csv_log_period,
-            file: file_obj,
+            writer: BufWriter::with_capacity(65536, file), // 64KB buffer - optimal for most filesystems
```
Dillonmcewan, line 38: This change makes a ton of sense

pictorus-internal/src/loggers/csv_logger.rs
```diff
@@ -14,33 +14,123 @@ use super::Logger;
...
             output_path,
             app_start_epoch: Duration::from_micros(
                 Utc::now()
                     .timestamp_micros()
                     .try_into()
                     .expect("Could not cast app start epoch as u64"),
             ),
+            header_written: false,
```
Dillonmcewan, line 46: Can't this already be inferred with `last_csv_log_time`?

pictorus-internal/src/loggers/csv_logger.rs
```diff
@@ -14,33 +14,123 @@ use super::Logger;
...
+                                        write!(self.buffer, "{n}").ok();
+                                    }
+                                    first_inner = false;
+                                }
+                                self.buffer.push(']');
+                            }
+                            first_elem = false;
+                        }
+                        self.buffer.push(']');
+                        self.buffer.push('"');
```
Dillonmcewan, line 124: Are we sure this is actually faster? I think the `serde_json::to_value` step has already done a lot of the heavy lifting at this point, so I'd be surprised if our own custom JSON serializer was actually much of an improvement

pictorus-internal/src/loggers/csv_logger.rs
```diff
@@ -49,28 +139,30 @@ impl Logger for CsvLogger {
...
+impl Drop for CsvLogger {
+    fn drop(&mut self) {
+        // Ensure buffer is flushed on drop
+        self.writer.flush().ok();
+    }
+}
+
 /// Formats the header for CSV output based on the provided data.
 /// This function extracts the field names from the data and formats them as a CSV header.
 pub fn format_header_csv(data: &impl serde::Serialize) -> String {
```
Dillonmcewan, line 164: Is this still used?

pictorus-internal/src/loggers/csv_logger.rs
```diff
@@ -86,9 +178,8 @@ pub fn format_header_csv(data: &impl serde::Serialize) -> String {
     header
 }
 
-/// Formats the samples for CSV output based on the provided data.
 pub fn format_samples_csv(data: &impl serde::Serialize) -> String {
```
Dillonmcewan, line 181: Is this still used?

pictorus-internal/src/loggers/csv_logger.rs
```diff
@@ -220,4 +343,204 @@ mod tests {
...
+        logger.log(&large_data, Duration::ZERO);
+        let elapsed = start.elapsed();
+
+        drop(logger);
+
+        // Should complete quickly even with large array
+        assert!(
+            elapsed < Duration::from_millis(10),
+            "Large array logging took too long: {elapsed:?}"
+        );
```
Dillonmcewan, line 512: I'm not sure this belongs in a unit test. These kind of perf comparisons are highly dependent on the platform and state of the system. It could be worth setting up some kind of benchmark suite for this, but even then we would want to make sure it's run in a semi-controlled environment

Dillonmcewan (commented):
Huge improvement! I think we should isolate some of these changes to see which ones are actually worth making, but overall I like the idea of the buffered file writing

PR #26, opened 2025-08-12
Updated lifetimes as needed indicated by clippy and rustc 1.89

amcelroy:


jscatena88 (approved):
Whew, thanks for tackling this for us. LGTM!

PR #28, opened 2025-09-08
Add px4 workspace older embassy

amcelroy:
Adds @jscatena88's px4-workspace to pictorus-rs. Not sure if we want to merge now and update later? I also had to branch from the previous commit, as the new Embassy upgrades were messing up the templates.

Either now or Future work:
- [ ] Generate new Rust ToPassType and FromPassType for all messages we wish to implement.
- [ ] Make sure implementations match uorb_dict.json in the Pictorus repo


amcelroy:
> When you mention having to branch off the older embassy version what kind of issues were you running into? What will the impact be?

The current codegen is using the older Embassy version, which is commit `024b1fa7890779325c6a848a616d12ee82c80d3e` of this repo. Branching from HEAD causes cargo version errors when generating code until the Embassy upgrade goes through.

jscatena88:
I hadn't noticed at first that both of the px4 crates were brought in under the px4-workspace directory. Is there a downside to keeping the workspace directory structure? I think generally keeping it flat would be preferable if we can swing it

px4-workspace/px4-msgs-sys/src/orb.rs
```diff
@@ -0,0 +1,156 @@
...
+//     ["Offset of field: orb_metadata::o_size"]
+//         [::core::mem::offset_of!(orb_metadata, o_size) - 4usize];
+//     ["Offset of field: orb_metadata::o_size_no_padding"]
+//         [::core::mem::offset_of!(orb_metadata, o_size_no_padding) - 6usize];
+//     ["Offset of field: orb_metadata::message_hash"]
+//         [::core::mem::offset_of!(orb_metadata, message_hash) - 8usize];
+//     ["Offset of field: orb_metadata::o_id"][::core::mem::offset_of!(orb_metadata, o_id) - 12usize];
+//     ["Offset of field: orb_metadata::o_queue"]
+//         [::core::mem::offset_of!(orb_metadata, o_queue) - 14usize];
+// };
```
jscatena88, line 41: This was an alignment test generated by bindgen that was failing when you switched between targets. We can probably just remove it

px4-workspace/px4-msgs-sys/c_api/uORB/topics/ackermann_velocity_setpoint.h
jscatena88, line 1: I had originally hoped to generate the rust bindings from these files at build time but ended up having to manually massage the bindgen output quite a bit so for now the `orb.rs` and `message_defs.rs` are just static. I wonder if in light of that if it is still worth  having all these header files in here. On the one hand it is nice for posterity but if we aren't generating off them they add a lot of noise to the repo and will probably end up going out of sync over time anyway. Maybe it'd be better to just add a note to the docs of this crate or those two files saying what git tag/hash of the PX4 repo we used as the source of them and leave it at that?

px4-workspace/pictorus-px4/src/ffi_protocol.rs
```diff
@@ -0,0 +1,1132 @@
...
+/// // Set up message subscriptions
+/// protocol.subscribe_to_message(SensorAccel::default());
+/// ```
+pub struct FfiProtocol {
+    /// Input messages that C++ writes and Rust reads
+    input_messages: Vec<MessageEntry>,
+    /// Output messages that Rust writes and C++ reads  
+    output_messages: Vec<MessageEntry>,
+}
+
```
jscatena88, line 82: I have vague dreams of using an intrusive linked list where each instance of an FfiOutputBlock/FfiInputBlock has statically allocated memory for these. This would allow us to drop the `alloc` requirement. Definitely low priority but I just wanted to put that idea down for posterity.

px4-workspace/module-shim/src/modules/pictorus_module/pictorus_module.h
```diff
@@ -0,0 +1,57 @@
...
+	 * @param force for a parameter update
+	 */
+	void parameters_update(bool force = false);
+
+
+	DEFINE_PARAMETERS(
+		(ParamFloat<px4::params::PICT_P_ROLL>) _pict_p_roll,
+		(ParamFloat<px4::params::PICT_I_ROLL>) _pict_i_roll,
+		(ParamFloat<px4::params::PICT_D_ROLL>) _pict_d_roll
+	)
```
jscatena88, line 52: When we add support for parameters this will also need to be generated

px4-workspace/px4-msgs-sys/c_api/uORB/topics/ackermann_velocity_setpoint.h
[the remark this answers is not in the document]
jscatena88: It should be assumed to be tied to a specific version. PX4 makes no promises about message stability in general. They have started to introduce the concept of versioned messages but only very very recently: https://docs.px4.io/main/en/middleware/uorb#message-versioning

That said I think a lot of the most common messages have been fairly stable so there is a decent chance it could work, but I wouldn't make any guarantees about any of this working on versions besides what it was built to. (even some other stuff with the uorb API and module defintion yaml file are fairly new in PX4 versions from what I understand)

amcelroy: Do you remember how much you had to massage? It seems like the safest course would be to lock in the uorb message files and generate the Rust bindings and dictionary for the codegen so we at least have some stable copy to generate against. 

As you noted above, a lot of people depend on these messages so there is probably stability and a deprecation path for messages, but we would have to follow the PX4 repo to catch any differences. That said, I've changed the codegen dictionary to be generated from the PX4-repo similar to how the chip configs are pulled in.

PR #30, opened 2025-09-12
Add all PX4 message impls

Dillonmcewan:
Add autogenerated PX4 message impls. This does not include messages with > 8 fields or messages with string fields


amcelroy (approved):
Do we need to remove all messages with c-strings or can we just remove those fields? Otherwise looks good with some spot checking of messages compiling to libraries.

Edit: Could we subscribe to the ones with strings (and not copy the strings) and not publish to those messages? WOuld require differentiating the messages, might not be worth it.

PR #32, opened 2025-09-15
Adds NoOpInputBlock and NoOpOutputBlock

amcelroy:
Adds special blocks that can be used as stubs when simulating input and output blocks. In the case of an input block, 0 values are returns for scalar and matrix types. For the output block, no operations are performed on the data.

This is helpful when simulating virtually generated input and output blocks that rely on some backend processing to size correctly.

### DIFF
--- a/pictorus-blocks/src/core_blocks/noop_input_block.rs
+++ b/pictorus-blocks/src/core_blocks/noop_input_block.rs
@@ -1,4 +1,4 @@
-use pictorus_traits::{Context, InputBlock, Pass, PassBy, Scalar};
+use pictorus_traits::{Context, InputBlock, Pass, PassBy};
 
 pub struct NoOpInputBlockParameters {}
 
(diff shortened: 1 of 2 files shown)


pictorus-blocks/src/core_blocks/noop_input_block.rs
```diff
@@ -0,0 +1,562 @@
...
+     }
+}
+
+/// This block is a special case for input blocks that need to be simulated
+/// but don't have a fixed message type, for example the uORB input block, which
+/// has an output size based on which message is selected. Output data is the
+/// default() value for the type.
+pub struct NoOpInputBlock<T> {
+    _phantom: core::marker::PhantomData<T>,
+    store: Option<T>,
```
Dillonmcewan, line 23: Agree with copilot. Does this actually need to be an option? I think we can just enforce `T: Default`

pictorus-blocks/src/core_blocks/noop_input_block.rs
```diff
@@ -0,0 +1,562 @@
...
+    type Output = f32;
+
+    fn input(
+        &mut self,
+        _parameters: &Self::Parameters,
+        _context: &dyn Context,
+    ) -> PassBy<'_, Self::Output> {
+        0.0
+    }
+}
```
Dillonmcewan, line 62: Can we do a blanket impl on `Scalar` here?

pictorus-blocks/src/core_blocks/noop_output_block.rs
```diff
@@ -0,0 +1,454 @@
...
+    type Inputs = f32;
+
+    fn output(
+        &mut self,
+        _parameters: &Self::Parameters,
+        _context: &dyn pictorus_traits::Context,
+        _inputs: pictorus_traits::PassBy<'_, Self::Inputs>,
+    ) {
+    }
+}
```
Dillonmcewan, line 73: Impl for `T: Scalar`?

Dillonmcewan (commented):
Couple questions on making the impls more generic to scalar types

pictorus-blocks/src/core_blocks/noop_output_block.rs
amcelroy: Thanks for the check, I made that entirely too complicated...

Dillonmcewan (approved):
So much simpler! Love it

PR #33, opened 2025-09-17
Added is_truthy to Matrix

amcelroy:
Adds `is_truthy()` to Matrix; used in cases where a Matrix needs to be checked if it is "true" or "false". If all values are T::default(), the matrix is "false", otherwise "true".

### DIFF
--- a/pictorus-traits/src/lib.rs
+++ b/pictorus-traits/src/lib.rs
@@ -575,14 +575,8 @@ where
     }
 
     pub fn is_truthy(&self) -> bool {
-        for col in 0..NCOLS {
-            for row in 0..NROWS {
-                if self.data[col][row] != T::default() {
-                    return true;
-                }
-            }
-        }
-        false
+        let default = T::default();
+        self.data.iter().flatten().any(|x| *x != default)
     }
 }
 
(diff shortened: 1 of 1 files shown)


pictorus-traits/src/lib.rs
```diff
@@ -573,6 +573,11 @@ where
             data: unsafe { mem::zeroed() },
         }
     }
+
+    pub fn is_truthy(&self) -> bool {
+        let default = T::default();
+        self.data.iter().flatten().any(|x| *x != default)
+    }
```
Dillonmcewan, line 580: What happens if you try to pass bytes into something that calls `is_truthy`? Will we hit a similar issue?

Dillonmcewan (approved):
LGTM with one question

pictorus-traits/src/lib.rs
amcelroy: The existing checks in `willAcceptNewConnection` seem to do a good job of filtering bytes. I'm not sure it is possible to add is_truthy to a ByteSliceSignal since it is a zero size type.  

Screenshot 2025-09-17 at 4 34 11 PM

PR #36, opened 2025-09-30
Add manual impl of Manual Control Setpoint

amcelroy:
Adds manual implementations for the `manual_control_setpoint_s` message in `pictorus-px4/src/message_impls.rs`.


Dillonmcewan (approved):
LGTM. This seems like a reasonable workaround until we can address the 8 input limit

PR #38, opened 2025-10-06
Adde ToScalar trait and impl for 1x1 matrix

amcelroy:
- Added ToScalar trait and impl for a 1x1 matrix

In some cases, we may have a 1x1 matrix that should be a scalar, this trait and impl allows for this case.

### DIFF
--- a/pictorus-traits/src/lib.rs
+++ b/pictorus-traits/src/lib.rs
@@ -561,6 +561,19 @@ where
     }
 }
 
+/// Convert struct into a scalar value
+pub trait ToScalar<S: Scalar> {
+    fn to_scalar(&self) -> S;
+}
+
+/// Convert a 1x1 matrix into a scalar value
+impl<S: Scalar> ToScalar<S> for Matrix<1, 1, S> {
+    #[inline]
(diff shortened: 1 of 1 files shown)


jscatena88:
Looks good, my only thoughts would be maybe we should implement the null case of `ToScalar` for a scalar? This would be similar to how Rust automatically offers `Into<T> for T`. That also got me thinking that maybe just a normal impl of `Into<>` would work here? Just a thought though, not married to it.

amcelroy:
I thought of Into as well, but was hoping to make the call more explicit so people didn't accidentally convert a matrix to a scalar, though maybe overthinking it. I can add a null case tomorrow.

jscatena88:
Yeah I'm probably overthinking it to. Maybe just go with the `ToScalar` for now. If we stuff it in `pictorus-blocks::traits` instead of `pictorus-traits` than that might be even lower stakes since that crate doesn't try to be nearly as "fully designed" and we can promote it to this crate later if we get more confidence. Or just merge it as is, we're definitely yak shaving here

jscatena88 (approved):

PR #39, opened 2025-10-07
Add macro for implementing the Pass trait out to 16 elements.

### LINKED ISSUE #31: Add support for blocks with > 8 I/O
We have encountered a few external protocols that commonly need to support > 8 inputs/outputs (CAN, FMU, uORB). We should have a way to support larger number of I/O with our trait system. The scope of this work does not actually involve implementing more variants of the existing core blocks, but just extending the number supported by our trait system for blocks that create dynamic defs. Ideally this is done in a way that makes it reasonably easy to change the number in the future (macros/build scripts?).

amcelroy:
Closes #31 

- Adds macro impl_pass_for_tuples which allows easier addition of pass traits beyond 8
- Adds impl_tuple_array_interop to do a similar task for the TupleEquivalent, within Rust limitations
- Linter removing a space in message_impls

Screenshot 2025-10-07 at 11 09 56 AM

### DIFF
--- a/pictorus-px4/src/message_impls.rs
+++ b/pictorus-px4/src/message_impls.rs
@@ -1519,7 +1519,7 @@ define_topics! {
 // Manual message impls below here
 // --------------------------------------------------------------------------------
 
-/// Manual Implementation for Manual Control Setpoint; Manual Control Setpoint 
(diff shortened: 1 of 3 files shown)


jscatena88 (approved):
LGTM! Nice job on the macros

Dillonmcewan (approved):
LGTM

PR #40, opened 2026-04-27
Change `BlockDef` signature 

amcelroy:
Changes:
- Updated `BlockDef` run signature to take: 
    -  immutable reference to an array of references that are of type BlockDataRead 
    - A mutable reference to an array of mutable references that are of type  BlockDataWrite
- Implement BlockDataWrite for: 
    - f64 (currently impl for &f64)
    - Matrix (currently impl for &Matrix)

### DIFF
--- a/pictorus-internal/src/utils.rs
+++ b/pictorus-internal/src/utils.rs
@@ -113,6 +113,7 @@ cfg_if::cfg_if! {
         use std::format;
         use std::io::Write;
         use std::prelude::rust_2021::*;
+        use std::panic;
 
         use log::{info, warn, LevelFilter};
         use chrono::Local;
--- a/pictorus-traits/src/custom_blocks.rs
+++ b/pictorus-traits/src/custom_blocks.rs
@@ -198,8 +198,7 @@ where
     }
 }
 
-impl<const NROWS: usize, const NCOLS: usize, T: Scalar> BlockDataWrite 
-    for Matrix<NROWS, NCOLS, T>
+impl<const NROWS: usize, const NCOLS: usize, T: Scalar> BlockDataWrite for Matrix<NROWS, NCOLS, T>
 where
     for<'a> &'a mut T: BlockDataWrite,
 {
--- a/pictorus-traits/src/custom_blocks.rs
+++ b/pictorus-traits/src/custom_blocks.rs
@@ -88,6 +88,12 @@ pub trait BlockDef {
     fn cleanup(&mut self) {}
 }
 
+// Returns a 'static [f64] of length 1 backing the bool's f64 representation.
+fn bool_as_matrix_data(b: bool) -> &'static [f64] {
+    static LUT: [f64; 2] = [0.0, 1.0];
+    core::slice::from_ref(&LUT[b as usize])
+}
+
 impl BlockDataRead for &bool {
     fn get_scalar(&self) -> f64 {
         if **self {
@@ -98,7 +104,7 @@ impl BlockDataRead for &bool {
     }
 
     fn get_matrix(&self) -> (usize, usize, &[f64]) {
-        unimplemented!("Cannot get matrix of scalar bool value")
+        (1, 1, bool_as_matrix_data(**self))
     }
 }
 
@@ -112,7 +118,7 @@ impl BlockDataRead for bool {
     }
 
     fn get_matrix(&self) -> (usize, usize, &[f64]) {
-        unimplemented!("Cannot get matrix of scalar bool value")
+        (1, 1, bool_as_matrix_data(*self))
     }
 }
 
@@ -124,7 +130,10 @@ macro_rules! scalar_block_data_read_impl {
                         (**self).into()
                     }
                     fn get_matrix(&self) -> (usize, usize, &[f64]) {
-                        unimplemented!("Can not get matrix of scalar {} value", stringify!($t))
+                        unimplemented!(
+                            "get_matrix on scalar {} is not supported; only f64 and bool scalars convert to 1x1 matrices",
+                            stringify!($t)
+                        )
                     }
                 }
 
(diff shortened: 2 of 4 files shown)


pictorus-internal/src/utils.rs
```diff
@@ -113,6 +113,7 @@ cfg_if::cfg_if! {
         use std::format;
         use std::io::Write;
         use std::prelude::rust_2021::*;
+        use std::panic;
```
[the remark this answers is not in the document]
amcelroy: This was put here for a clippy warning, panic! macros used without an import will cause an issue in the future.

pictorus-traits/src/custom_blocks.rs
```diff
@@ -102,6 +102,20 @@ impl BlockDataRead for &bool {
...
+        if *self {
+            1.0
+        } else {
+            0.0
+        }
+    }
+
+    fn get_matrix(&self) -> (usize, usize, &[f64]) {
+        unimplemented!("Cannot get matrix of scalar bool value")
+    }
```
Dillonmcewan, line 127: I think the existing `BlockData` equivalent returns a 1x1 matrix in this case. If I remember correctly we allow you treat a scalar as a matrix, but not vice-versa. We should decide if we want to drop support for this

Dillonmcewan (commented):
It looks like this PR changes some of the assumptions around scalar/matrix interoperability. Not opposed to this necessarily, but might be worth a discussion to decide if we're okay with that change

pictorus-traits/src/custom_blocks.rs
amcelroy: That is a good call out, ~~let me update this PR to at least have that behavior while we discuss. Adding `try_get_matrix` and `try_get_scalar` that return an `Option<...>` might be another approach that people can use that would be safer.~~

Taking a second look, we might not be able to to this, since we pre-allocate either a matrix or an f64 to pass in as a mutable reference as the output location.

jscatena88: I think we could use a little static trick to do this:

```rust
    fn get_matrix(&self) -> (usize, usize, &[f64]) {
        static TRUE_VAL:  f64 = 1.0;
        static FALSE_VAL: f64 = 0.0;

        let slice = if *self {
            std::slice::from_ref(&TRUE_VAL)
        } else {
            std::slice::from_ref(&FALSE_VAL)
        };
        (1, 1, slice)
    }
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=333267a9b6f80c302520483ccc1d11d5

amcelroy: Since we are reading references from an immutable static value, that should work. Nice trick! 

Reworking bool and f64 scalar-to-matrix would look something like this patch
```
diff --git a/pictorus-traits/src/custom_blocks.rs b/pictorus-traits/src/custom_blocks.rs
index 2fd77c6..ce9cc01 100644
--- a/pictorus-traits/src/custom_blocks.rs
+++ b/pictorus-traits/src/custom_blocks.rs
@@ -88,6 +88,14 @@ pub trait BlockDef {
     fn cleanup(&mut self) {}
 }
 
+// Returns a 'static [f64] of length 1 backing the bool's f64 representation.
+fn bool_as_matrix_data(b: bool) -> &'static [f64] {
+    static LUT: [f64; 2] = [0.0, 1.0];
+    core::slice::from_ref(&LUT[b as usize])
+}
+
 impl BlockDataRead for &bool {
     fn get_scalar(&self) -> f64 {
         if **self {
@@ -98,7 +106,7 @@ impl BlockDataRead for &bool {
     }
 
     fn get_matrix(&self) -> (usize, usize, &[f64]) {
-        unimplemented!("Cannot get matrix of scalar bool value")
+        (1, 1, bool_as_matrix_data(**self))
     }
 }
 
@@ -112,7 +120,7 @@ impl BlockDataRead for bool {
     }
 
     fn get_matrix(&self) -> (usize, usize, &[f64]) {
-        unimplemented!("Cannot get matrix of scalar bool value")
+        (1, 1, bool_as_matrix_data(*self))
     }
 }
 
@@ -124,7 +132,10 @@ macro_rules! scalar_block_data_read_impl {
                         (**self).into()
                     }
                     fn get_matrix(&self) -> (usize, usize, &[f64]) {
-                        unimplemented!("Can not get matrix of scalar {} value", stringify!($t))
+                        unimplemented!(
+                            "get_matrix on scalar {} is not supported; only f64 and bool scalars convert to 1x1 matrices",
+                            stringify!($t)
+                        )
                     }
                 }
 
@@ -133,14 +144,37 @@ macro_rules! scalar_block_data_read_impl {
                         (*self).into()
                     }
                     fn get_matrix(&self) -> (usize, usize, &[f64]) {
-                        unimplemented!("Can not get matrix of scalar {} value", stringify!($t))
+                        unimplemented!(
+                            "get_matrix on scalar {} is not supported; only f64 and bool scalars convert to 1x1 matrices",
+                            stringify!($t)
+                        )
                     }
                 }
             )+
         };
     }
 
-scalar_block_data_read_impl!(u8, i8, u16, i16, u32, i32, f32, f64);
+scalar_block_data_read_impl!(u8, i8, u16, i16, u32, i32, f32);
+
+impl BlockDataRead for &f64 {
+    fn get_scalar(&self) -> f64 {
+        **self
+    }
+
+    fn get_matrix(&self) -> (usize, usize, &[f64]) {
+        (1, 1, core::slice::from_ref(*self))
+    }
+}
+
+impl BlockDataRead for f64 {
+    fn get_scalar(&self) -> f64 {
+        *self
+    }
+
+    fn get_matrix(&self) -> (usize, usize, &[f64]) {
+        (1, 1, core::slice::from_ref(self))
+    }
+}
```

with some additional unit tests. We just need to pull the f64 impl out of the macro and use Jason trick for bools. Might be worth another discussion about non-f64 types and conversion blocks once we have BlockData removed?

pictorus-traits/src/custom_blocks.rs
```diff
@@ -88,6 +88,12 @@ pub trait BlockDef {
     fn cleanup(&mut self) {}
 }
 
+// Returns a 'static [f64] of length 1 backing the bool's f64 representation.
+fn bool_as_matrix_data(b: bool) -> &'static [f64] {
+    static LUT: [f64; 2] = [0.0, 1.0];
+    core::slice::from_ref(&LUT[b as usize])
```
jscatena88: super nit: `as` casts always give me the heebie-jeebies, especially if they can be avoided. Could we switch this to using an if/else or match statement?

jscatena88 (approved):
LGTM! just one optional style nit

pictorus-traits/src/custom_blocks.rs
amcelroy: Not a problem at all

PR #43, opened 2026-05-12
Remove BlockData from params and ICs

jscatena88:


pictorus-internal/src/utils.rs
```diff
@@ -145,15 +145,100 @@ cfg_if::cfg_if! {
...
         }
 
-        impl LoadableParams for BlockData {
-            fn parse(source: &str, default: Option<&Self>) -> Option<Self> {
-                let parsed_val: Vec<f64> = string_to_vec::<f64>(source);
+        impl<const N: usize> LoadableParams for [u8; N] {
+            fn parse(source: &str, _default: Option<&Self>) -> Option<Self> {
+                let mut buffer = [0u8; N];
+                let mut input_len = 0;
+               let trimmed = source.trim().trim_start_matches('[').trim_end_matches(']');
```
amcelroy: Nit, not aligned.

pictorus-block-data/src/lib.rs
```diff
@@ -859,6 +859,8 @@ impl ToPass<bool> for BlockData {
 
 impl<const N: usize> ToPass<[u8; N]> for BlockData {
     fn to_pass(&self) -> [u8; N] {
+        // This is doing a lot of heavy lifting!
+        //If an undersized Vec is passed in, this will just take the first N elements of the data.
```
amcelroy, line 863: It could also overflow if ._data is bigger than N, we should use `.take(N)` for this one and `ToPass<[f64; N]>`

pictorus-internal/src/utils.rs
```diff
@@ -145,15 +145,100 @@ cfg_if::cfg_if! {
...
+                for (i, byte) in trimmed.split(&[',']).enumerate() {
```
amcelroy: Can we use .take(N).enumerate() here as well to avoid the i >= N check? Same for the other `impl` with a `break` check

pictorus-blocks/src/matrix_ext.rs
```diff
@@ -38,3 +38,40 @@ where
...
+    for Matrix<NROWS, NCOLS, T>
+{
+    type Elem = T;
+
+    fn from_flat_array(rows: usize, cols: usize, data: &[Self::Elem]) -> Self {
+        // This accepts data as row-major, but pictorus_traits::Matrix is column-major, so we need to transpose the data
+        let mut m = Self::zeroed();
+        for r in 0..rows {
+            for c in 0..cols {
+                m.data[c][r] = data[r * cols + c];
```
amcelroy: Should we do a check against rows / cols against NROWS / NCOLS here in case `data` has a different number of elements? In one case we would fill with 0's, in the other case we would overflow.

pictorus-internal/src/utils.rs
```diff
@@ -145,15 +145,100 @@ cfg_if::cfg_if! {
...
+            fn parse(source: &str, _default: Option<&Self>) -> Option<Self> {
+                let mut matrix = Matrix::zeroed();
+                let mut input_len = 0;
+
+                // Strip outer brackets and whitespace
+                let trimmed = source.trim()
+                    .trim_start_matches('[')
+                    .trim_end_matches(']');
+                // Split into rows on "],["
+               for (row_i, row_str) in trimmed.split("],").enumerate() {
```
amcelroy: Nit, alignment.

pictorus-internal/src/utils.rs
```diff
@@ -145,15 +145,100 @@ cfg_if::cfg_if! {
...
+               let trimmed = source.trim().trim_start_matches('[').trim_end_matches(']');
+                for (i, byte) in trimmed.split(&[',']).enumerate() {
+                    if i >= N {
+                        break;
+                    }
+                    if let Ok(parsed_byte) = byte.trim().parse::<u8>() {
+                        buffer[i] = parsed_byte;
+                        input_len = i + 1;
+                    } else {
+                        return None; // Parsing failed
```
amcelroy: Should we throw up `PictorusError` here similar to the UDP parse failure? Applies to the 2 other `impl`. 

Edit: It looks like we only affirm variable parsing success in `load_param`

amcelroy (approved):
Looks pretty good! Excited to get BlockData out of params / ICs for the follow up work for feedback loops.

pictorus-blocks/src/matrix_ext.rs
jscatena88: Yeah, it is probably worth doing something smarter than silently leaving zeros or letting Rust panic on out of bounds access. 

We don't have much in the way of error handling strategy in `pictorus-blocks`. Maybe I'll just check if `data.len() == N` and if not panic with a descriptive error message?

pictorus-internal/src/utils.rs
jscatena88: I actually went in and refactored these this morning to use `serde_json` where ever possible. Rather than hand rolling our own parsing logic. It was already a dependency so it makes sense to use it I think. I'll be pushing that shortly

pictorus-internal/src/utils.rs
jscatena88: I just reworked the logic of load_param to at least give a warning if a variable with the right name is found in Env or diagram_params.json but couldn't be parsed

pictorus-internal/src/utils.rs
jscatena88: This ended up being because of the `cfg_if` macro wrapper. It was preventing cargo from formatting it. I reworked it to avoid the macro (and was able to remove `cfg_if` as a dependency 🥳 ). So hopefully the formatting should be good from now on

pictorus-blocks/src/matrix_ext.rs
```diff
@@ -52,12 +52,19 @@ impl<const NROWS: usize, const NCOLS: usize, T: pictorus_traits::Scalar> MatrixE
...
+            );
+        }
         // This accepts data as row-major, but pictorus_traits::Matrix is column-major, so we need to transpose the data
         let mut m = Self::zeroed();
-        for r in 0..rows {
-            for c in 0..cols {
-                m.data[c][r] = data[r * cols + c];
+        for r in 0..NROWS {
+            for c in 0..NCOLS {
+                m.data[c][r] = data[r * NCOLS + c];
```
amcelroy, line 67: Perfect!

pictorus-internal/src/utils.rs
```diff
@@ -169,15 +169,19 @@ mod std_utils {
                 return Some([parsed; N]);
             }
             // Fall back to taking the string as bytes
-            if source.len() != N {
+            // We need to handle `/x**` hex format and turn it into raw bytes
+            let escaped_bytes = smashquote::unescape_bytes(source.as_bytes())
```
amcelroy, line 173: 🎉

PR #44, opened 2026-05-13
Fix for serde bounds of large arrays

jscatena88:


pictorus-internal/src/utils.rs
```diff
@@ -120,6 +121,15 @@ mod std_utils {
     use log::{LevelFilter, info, warn};
 
     pub type DiagramParams = HashMap<String, HashMap<String, String>>;
+    #[derive(serde::Deserialize)]
+    pub struct BigArrayWrap<const N: usize, T: DeserializeOwned>(
+        #[serde(with = "serde_big_array::BigArray")] [T; N],
```
amcelroy, line 126: Interesting syntax, just just skips a named field and you can access it below using .0, nifty. Didn't know you could stack the macro in there.

amcelroy (approved):
Seems like a good way to bypass the Serde 32 element limitation.

Dillonmcewan (approved):

PR #45, opened 2026-05-14
Add .buffer() trait function to allow a way to get access to a blocks last run value.

amcelroy:
TL; DR:
- Add .buffer() function to Generator and Process blocks.
- Implement the .buffer() function on all generator and process blocks.
- Create a valid buffer field which can be used to replace .data access and caching.
- Warn users if they use a ::default() constructor for blocks with `HasIc`

Longer winded reason for changes:

Feedback loops necessitate that any block at a feedback edge be able to reach into an upstream block before the upstream block has run and grab a valid value. This was previously done using the heap allocated .data field backed by `OldBlockData`, which we are trying to phase out.

For example, a sum block with inputs from a constant block and a delay delay:
Screenshot 2026-05-14 at 8 56 03 AM

Produces:
```
impl MainState {
    pub fn new(_context: &Context) -> Self {
        // Constant1
        let constant1 = ConstantBlock::default();

        // Delay1
        let delay1 = DelayBlock::default();

        // Sum1
        let sum1 = SumBlock::default();

        MainState {
            last_time_s: -1.0,
            constant1,
            delay1,
            sum1,
        }
    }

    pub fn run(&mut self, context: &mut Context) {
        let app_time_s = context.app_time_s();
        let runtime_ctx = context.get_runtime_context();

        if self.last_time_s == -1.0 {
            self.last_time_s = app_time_s;
        }
        let timestep_s: f64 = app_time_s - self.last_time_s;

        log::debug!(
            "-- State::MainState iteration. Time: {}s (dt: {}s) ",
            app_time_s,
            timestep_s
        );

        // Constant1
        let constant1_0 = self.constant1.generate(
            &context.application_parameters.constant1_a65d6,
            &runtime_ctx,
        );

        // Delay1
        let delay1_0 = self.delay1.process(
            &context.application_parameters.delay1_a65d9,
            &runtime_ctx,
            self.sum1.data.to_pass(),
        );
        context.log_data.delay1_a65d9_0(delay1_0);
        // Sum1
        let sum1_0 = self.sum1.process(
            &context.application_parameters.sum1_a65d7,
            &runtime_ctx,
            (constant1_0, delay1_0),
        );
        context.log_data.sum1_a65d7_0(sum1_0);

        self.last_time_s = app_time_s;
    }

    pub fn post_run(&mut self) {}
}
```

The Delay block uses a value from the Sum block BEFORE the Sum block has been called. In most cases, this defaults to `T::default()` values. In a world without `BlockData`, we need a similar mechanism.

This PR bridges the gap, by forcing all `ProcessBlock` and `GeneratorBlock`s to implement a .buffer() function that returns a buffer value (preferably in a stack allocated field) for a block even before it has been run. Most blocks were pretty easy since the generics have been constrained to require the `Default` trait which was used to fill out `OldBlockData` at construction. This required blocks to have their `buffer: Option<T::Output>` converted to `buffer: T::Output` and a value of `T::default()` assigned in the constructor. Now the buffer (or store or output) field provides the same service as .data.

Several blocks had internal traits (e.g. `Apply`) and implementations that passed in an Option<T> and were updated to use the updated non-option buffer. `AggregateBlock` is an example of this.

Finally, there are 8 stateful blocks that have Initial Conditions, which are passed in via `Parameters`, which the buffer should reflect. Codegen orders block execution in a way that a stateful block will go first; in the example above, Delay executes before Sum. We call `::default()` when constructing blocks, which doesn't correctly seed either `.data` or `.buffer` fields. If the user were to hand coded blocks in such a way that they pulled `delay1.buffer()` BEFORE `delay1.process()` was run, they would get the values set in `::default()` and not the initial conditions. The correct action should be to instantiate an IC block using the `HasIc::new(...)` function so that the `.data` and `.buffer` values are correctly seeded with the IC. This is why there is a `warn!` in the `::default()` function implementation for any block that implements `HasIc`.

After all of this is said and done, the new `.data` replacement looks like:
```
impl MainState {
    pub fn new(context: &Context) -> Self {
        // Constant1
        let constant1 = ConstantBlock::default();

        // Delay1
        let delay1 = DelayBlock::new(&context.application_parameters.delay1_4d2b7);

        // Sum1
        let sum1 = SumBlock::default();

        MainState {
            last_time_s: -1.0,
            constant1,
            delay1,
            sum1,
        }
    }

    pub fn run(&mut self, context: &mut Context) {
        let app_time_s = context.app_time_s();
        let runtime_ctx = context.get_runtime_context();

        if self.last_time_s == -1.0 {
            self.last_time_s = app_time_s;
        }
        let timestep_s: f64 = app_time_s - self.last_time_s;

        log::debug!(
            "-- State::MainState iteration. Time: {}s (dt: {}s) ",
            app_time_s,
            timestep_s
        );

        // Constant1
        let constant1_0 = self.constant1.generate(
            &context.application_parameters.constant1_4d2b4,
            &runtime_ctx,
        );

        // If the user tried to grab delay1.data after using the `::default()` constructor, 
        // it would be incorrect here. Using ::new(...) avoids this issue.

        // Delay1
        let delay1_0 = self.delay1.process(
            &context.application_parameters.delay1_4d2b7,
            &runtime_ctx,
            self.sum1.buffer(),    // <--- no more .data
        );
        context.log_data.delay1_4d2b7_0(delay1_0);
        // Sum1
        let sum1_0 = self.sum1.process(
            &context.application_parameters.sum1_4d2b5,
            &runtime_ctx,
            (constant1_0, delay1_0),
        );
        context.log_data.sum1_4d2b5_0(sum1_0);

        self.last_time_s = app_time_s;
    }

    pub fn post_run(&mut self) {}
}
```

Additional things to consider:
- We might be able to go back and clean up some of the helper traits since we have a mutable place to store interim data.
- It may be worth removing the ::default() constructor for any block with `HasIc` instead of warning the user

### DIFF
--- a/pictorus-blocks/src/core_blocks/change_detection_block.rs
+++ b/pictorus-blocks/src/core_blocks/change_detection_block.rs
@@ -18,7 +18,7 @@ use strum::EnumString;
 /// so it will always emit `false`
 pub struct ChangeDetectionBlock<T: Apply> {
     pub data: OldBlockData,
-    buffer: Option<T::Output>,
+    buffer: T::Output,
     last_input: Option<T>,
 }
 
@@ -28,9 +28,14 @@ where
     OldBlockData: FromPass<T::Output>,
 {
     fn default() -> Self {
+        log::warn!(
+            "ChangeDetectionBlock constructed via Default; IC not seeded. \
+             Prefer ChangeDetectionBlock::new(&parameters) — otherwise buffer() will return \
+             T::default() until the first process() call."
+        );
         Self {
             data: <OldBlockData as FromPass<T::Output>>::from_pass(T::Output::default().as_by()),
-            buffer: None,
+            buffer: T::Output::default(),
             last_input: None,
         }
     }
@@ -43,7 +48,7 @@ where
 {
     fn new(parameters: &Self::Parameters) -> Self {
         ChangeDetectionBlock::<T> {
-            buffer: Some(T::Output::default()),
+            buffer: T::Output::default(),
             data: <OldBlockData as FromPass<T::Output>>::from_pass(T::Output::default().as_by()),
             last_input: Some(parameters.ic),
         }
@@ -69,13 +74,17 @@ where
         self.data = OldBlockData::from_pass(output);
         output
     }
+
+    fn buffer(&self) -> PassBy<'_, Self::Output> {
+        self.buffer.as_by()
+    }
 }
 
 pub trait Apply: Pass + Sized + Copy {
     type Output: Pass + Default;
 
     fn apply<'s>(
-        store: &'s mut Option<Self::Output>,
+        store: &'s mut Self::Output,
         input: PassBy<Self>,
         params: &Parameters<Self>,
         last_input: &mut Option<Self>,
@@ -86,7 +95,7 @@ impl<C: ChangeDetect> Apply for C {
     type Output = f64;
 
     fn apply<'s>(
-        store: &'s mut Option<Self::Output>,
+        store: &'s mut Self::Output,
         input: PassBy<Self>,
         params: &Parameters<Self>,
         last_input: &mut Option<Self>,
@@ -95,7 +104,7 @@ impl<C: ChangeDetect> Apply for C {
         let old_last_input = last_input.replace(input);
         let old_last_input = old_last_input.unwrap_or(params.ic);
         let res = Self::change_detect(input, old_last_input, params.change_mode);
-        *store = Some(res);
+        *store = res;
         res.as_by()
     }
 }
@@ -104,7 +113,7 @@ impl<const NROWS: usize, const NCOLS: usize, C: ChangeDetect> Apply for Matrix<N
     type Output = Matrix<NROWS, NCOLS, f64>;
 
     fn apply<'s>(
-        store: &'s mut Option<Self::Output>,
+        store: &'s mut Self::Output,
         input: PassBy<Self>,
         params: &Parameters<Self>,
         last_input: &mut Option<Self>,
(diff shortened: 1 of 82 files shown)


pictorus-blocks/src/core_blocks/change_detection_block.rs
```diff
@@ -28,9 +28,14 @@ where
     OldBlockData: FromPass<T::Output>,
 {
     fn default() -> Self {
+        log::warn!(
+            "ChangeDetectionBlock constructed via Default; IC not seeded. \
+             Prefer ChangeDetectionBlock::new(&parameters) — otherwise buffer() will return \
+             T::default() until the first process() call."
+        );
```
Dillonmcewan: Should we even allow this? Maybe we should add a static assert so calling this gets treated as a compiler error. Even better would be to remove the `Default` impl entirely for these blocks, but that would require a broader trait re-work

pictorus-blocks/src/core_blocks/delay_block.rs
```diff
@@ -34,6 +36,11 @@ where
     pictorus_block_data::BlockData: FromPass<T>,
 {
     fn default() -> Self {
+        log::warn!(
```
Dillonmcewan: Same question here/on the rest of the IC blocks. Maybe we just make the `default` route not valid for blocks with IC

Dillonmcewan (approved):
LGTM. As discussed in separate conversation, I think we should be able to drop the buffer requirement with a slight rework of feedback loop handling, but for now this moves us in the right direction for `BlockData`. Also looks like you added some warnings for `::default()` calls on blocks with IC, but I wonder if we should just disallow this entirely

PR #47, opened 2026-05-18
Remove ComponentInput block and all references

Dillonmcewan:
Removes component input block since it is no longer required by generated code. I also elected to remove the broken examples from the PX4 repo. They were both in a bad state without clear documentation on how to use them. If we add examples we need to make sure they are exercised in CI and documented properly


amcelroy (approved):
LGTM

PR #50, opened 2026-05-19
Remove BlockData from single-input/output Process blocks

amcelroy:
- Remove BlockData from Process blocks that are single input / single output
- Update unit tests

### DIFF
--- a/pictorus-blocks/src/core_blocks/clamp_block.rs
+++ b/pictorus-blocks/src/core_blocks/clamp_block.rs
@@ -166,17 +166,12 @@ mod test {
                     let input = Matrix {
                         data: [[pos_2.into(), $type::zero().into()], [$type::one().into(), neg_2.into()]]
                     };
+                    let expected = Matrix {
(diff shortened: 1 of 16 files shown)


pictorus-blocks/src/core_blocks/clamp_block.rs
```diff
@@ -187,10 +173,10 @@ mod test {
                             data: [[$type::one(), $type::zero()], [$type::one(), -$type::one()]]
                         }
                     );
-                    assert_eq!(
-                        block.data,
-                        OldBlockData::from_matrix(&[&[$type::one().into(), $type::one().into()], &[$type::zero().into(), (-$type::one()).into()]])
-                    );
+                    assert_eq!(block.buffer().data[0][0], $type::one());
```
Dillonmcewan: Is there a way to do these assertions in a less verbose way? Annoying yo have to call .buffer per index

Dillonmcewan (approved):
It looks like there's a couple spots where tests got more verbose. Otherwise lgtm

PR #51, opened 2026-05-20
Remove BlockData from remaining ProcessBlocks

amcelroy:
- IsValid::is_valid trait function now returns a bool instead of BlockData
- Removed "pub data: OldBlockData" from the rest of the process blocks.
- Updated unit tests

### DIFF
--- a/pictorus-blocks/src/core_blocks/bytes_unpack_block.rs
+++ b/pictorus-blocks/src/core_blocks/bytes_unpack_block.rs
@@ -2,17 +2,14 @@ extern crate alloc;
 use core::time::Duration;
 
 use crate::{traits::Scalar, IsValid};
-use alloc::vec::Vec;
 use generic_array::{ArrayLength, GenericArray};
 use typenum::{Const, NonZero, Sub1, ToUInt, B1, U};
 
 use crate::byte_data::{parse_byte_data_spec, try_unpack_data, ByteOrderSpec, DataType};
-use pictorus_block_data::BlockData as OldBlockData;
 use pictorus_traits::{ByteSliceSignal, PassBy, ProcessBlock};
 
 /// Unpacks a byte slice into a specified number of outputs based on the provided data types and byte order.
 pub struct BytesUnpackBlock<const N: usize> {
-    pub data: Vec<OldBlockData>,
     buffer: [f64; N],
     last_valid_time: Option<Duration>,
 }
@@ -20,12 +17,7 @@ pub struct BytesUnpackBlock<const N: usize> {
 impl<const N: usize> Default for BytesUnpackBlock<N> {
     fn default() -> Self {
         let buffer = [0.0; N];
-        let data = buffer
-            .iter()
-            .map(|_| OldBlockData::from_scalar(0.0))
-            .collect();
         BytesUnpackBlock {
-            data,
             buffer,
             last_valid_time: None,
         }
@@ -86,11 +78,6 @@ where
             // If we failed to unpack and it is stale, keep the old buffer but mark as invalid
             self.buffer[N - 1] = 0.0; // last element is is_valid flag
         }
-        self.data = self
-            .buffer
-            .iter()
-            .map(|&x| OldBlockData::from_scalar(x))
-            .collect();
         &self.buffer
     }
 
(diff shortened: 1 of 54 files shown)


pictorus-blocks/src/core_blocks/json_load_block.rs
```diff
@@ -1,11 +1,10 @@
 extern crate alloc;
 
 use alloc::string::String;
-use alloc::vec;
 use alloc::vec::Vec;
 
 use miniserde::json::{self, Array, Number, Object, Value};
-use pictorus_block_data::{BlockData as OldBlockData, BlockDataType, FromPass};
+use pictorus_block_data::BlockDataType;
```
amcelroy, line 7: This is the last reference to `pictorus_block_data` in `pictorus-blocks`.  There are notes in the `Parameters` struct and implementation that discusses an alternative. 

It seemed better to address this in a [follow-up ticket](https://github.com/Pictorus-Labs/pictorus/issues/3621) since this is about remove BlockData.

pictorus-blocks/src/lib.rs
```diff
@@ -52,7 +51,6 @@ pub struct ParseEnumError;
 /// Blocks that implement this trait should output a false value for `is_valid`
 /// if the output is not in a valid state.
 pub trait IsValid {
-    // This still uses the deprecated BlockData type.
-    // Eventually we will need to update/replace this trait when BlockData is removed.
-    fn is_valid(&self, app_time_s: f64) -> BlockData;
+    /// Returns `true` if the block output is in a valid state, `false` otherwise.
+    fn is_valid(&self, app_time_s: f64) -> bool;
```
amcelroy, line 55: Returning as `bool` doesn't seem to cause any issues in unit / integration tests and a quick test with comparison and switch blocks. We are converting to float in `value_accessor_components` is an if statement.

Image

Code snippets from the project above:
```
        let comparison1_0 = self.comparison1.process(
            &context.application_parameters.comparison1_9f061,
            &runtime_ctx,
            (
                if serial_receive1_is_valid { 1.0 } else { 0.0 },
                constant1_0,
            ),
        );
```

```
        let switch1_0 = self.switch1.process(
            &context.application_parameters.switch1_9f081,
            &runtime_ctx,
            (
                if matrix_inverse1_is_valid { 1.0 } else { 0.0 },
                &matrix_inverse1_0,
                &constant3_0,
            ),
        );
```

pictorus-blocks/src/core_blocks/bytes_unpack_block.rs
```diff
@@ -100,8 +87,8 @@ where
 }
 
 impl<const N: usize> IsValid for BytesUnpackBlock<N> {
-    fn is_valid(&self, _app_time_s: f64) -> OldBlockData {
-        OldBlockData::from_scalar(self.buffer[N - 1])
+    fn is_valid(&self, _app_time_s: f64) -> bool {
```
Dillonmcewan, line 90: I seem to remember there being one or 2 lingering issues with `bool` signals. Might be worth doing a quick check that these work as expected in a diagram

Dillonmcewan (approved):
LGTM! Huge win. One question about bool usage, but it looks like you already addressed in a comment

pictorus-blocks/src/core_blocks/bytes_unpack_block.rs
amcelroy: Screenshot 2026-05-20 at 1 25 34 PM

Expanded some of the blocks used, including the `bytes_unpack_block` and a vector merge of an is_valid flag. 
- For `bytes_unpack_block`, we are returning an array of floats and massage the `is_valid` bool -> float and it is treated as an f64 in codegen by access the N-1 array index.
- The vector merge of is_valid uses the same bool -> float trick to generate a vector of f64:
```
        // VectorMerge2
        let vector_merge2_0 = self.vector_merge2.process(
            &context.application_parameters.vector_merge2_75dc6,
            &runtime_ctx,
            (
                if serial_receive1_is_valid { 1.0 } else { 0.0 },
                if serial_receive1_is_valid { 1.0 } else { 0.0 },
            ),
        );
```

We should probably loop back on this in the future, however unit and integration tests are passing, and if we hit any issues we can always return an f64 from IsValid::is_valid if other fixes are too much trouble.

PR #54, opened 2026-05-28
Separate blocks into core, std, and alloc

amcelroy:
Now that `BlockData` is gone, the majority of blocks are `no_std` friendly and don't rely on `alloc`. There are still a handful of blocks that do rely on `alloc`; these blocks have been moved into the "alloc_blocks" module while we investigate plans for `alloc` free alternatives. 

- Add "alloc" feature flag to pictorus-blocks features. The "alloc" feature is a default feature for now so we don't need to change codegen.
- Gate Serialize trait behind the "alloc" feature.
- Gate certain ByteSliceSignal implementations that rely on Vec behind the "alloc" feature flag.
- Create a new "alloc_blocks" module that is gated behind the "alloc" feature.
- Move blocks that rely on alloc our of core_blocks
   - UDP blocks -> std_blocks. This required adding the feature opt in flag "std" to pictorus-sim and pictorus-linux crates.
   - Move any block that relies on "alloc" and is still a cross platform library into the "alloc_blocks" module.
- Added new CI and pre-commit checks to ensure no alloc functions are used in core blocks.
- New checks exposed that aggregate_block was using sort_by which is in the alloc library. Changed to sort_unstable_by which is in the core library.
- `CanReceiveBlock` had a vestigial extern crate alloc that was removed.


.github/workflows/ci-tests.yaml
```diff
@@ -29,10 +29,16 @@ jobs:
           components: rustfmt, clippy
       - name: Run Clippy
         run: cargo clippy --all-targets --all-features
+      - name: Run Clippy (pictorus-blocks, alloc only)
+        # Explicitly disable defaults so this exercises the alloc-without-std
+        # combo.
+        run: cargo clippy -p pictorus-blocks --all-targets --no-default-features --features alloc
```
Dillonmcewan: Wouldn't this already be covered by the `--all-features` flag in the above clippy command?

.github/workflows/ci-tests.yaml
```diff
@@ -29,10 +29,16 @@ jobs:
...
       - name: Run Cargo Deny
         uses: EmbarkStudios/cargo-deny-action@v2
       - name: Run all tests
         run: cargo test --all-features
+      - name: Run pictorus-blocks tests (alloc only)
+        run: cargo test -p pictorus-blocks --lib --no-default-features --features alloc
```
Dillonmcewan, line 40: Same here. Is this not redundant with the existing test command?

pictorus-blocks/src/lib.rs
```diff
@@ -19,19 +22,29 @@
 #[cfg(any(feature = "std", test))]
 extern crate std;
 
-// TODO: Remoove this when we no longer require alloc
+// `alloc` is needed when the `alloc` feature is on, when `std` is on (std implies alloc),
+// or when running tests (the test harness pulls in std, and several core_blocks tests
+// reach for alloc paths).
+#[cfg(any(feature = "alloc", feature = "std", test))]
```
Dillonmcewan: Can't we just have the `std` feature also enable the `alloc` feature directly?

pictorus-blocks/src/core_blocks/passthrough_block.rs
```diff
@@ -51,7 +51,9 @@ impl<T: DefaultStorage> ProcessBlock for PassthroughBlock<T> {
 #[cfg(test)]
 mod tests {
     use crate::testing::StubContext;
-    use pictorus_traits::{ByteSliceSignal, Matrix, Pass};
+    #[cfg(feature = "alloc")]
```
Dillonmcewan, line 54: Isn't `alloc` always enabled for tests? Is this feature check necessary?

pictorus-blocks/src/lib.rs
Dillonmcewan: I noticed a lot of the blocks also declare an `extern crate alloc`. Is that needed for each block if we have it handled at the module level?

pictorus-blocks/Cargo.toml
```diff
@@ -50,4 +50,8 @@ byteorder = { version = "1.5.0", features = ["std"] }
 derive-new = { version = "0.7.0", default-features = false }
 
 [features]
-std = ["dep:rustfft", "dep:chrono"]
+# TODO: `alloc` is a default feature for now to keep platform crates simple and to
+# match the current UI/codegen output. This should eventually be moved to opt-in.
+default = ["alloc"]
+alloc = []
```
Dillonmcewan, line 56: This should probably imply `pictorus-traits/alloc` as well

pictorus-blocks/Cargo.toml
```diff
@@ -50,4 +50,8 @@ byteorder = { version = "1.5.0", features = ["std"] }
...
+std = ["alloc", "dep:rustfft", "dep:chrono"]
```
Dillonmcewan, line 57: Oh, it looks like we already enable `alloc` when the `std` feature is enabled. I think that extra check to enable alloc in the module can be removed

.pre-commit-config.yaml
```diff
@@ -15,3 +15,12 @@ repos:
         entry: cargo clippy -- -D warnings
         language: system
         pass_filenames: false
+      - id: check-no-default-features
```
Dillonmcewan: This would get caught in CI, correct? Maybe not necessary in the pre-commit

Dillonmcewan (commented):
Seems like there might be some redundant checks/config happening, but otherwise looks good.

pictorus-blocks/src/lib.rs
amcelroy: We can remove `extern crate alloc` fromt he blocks, I was on the fence and left it there so we could do a quick audit of blocks, but as they are gated and live in a different module that seems redundant.

pictorus-blocks/Cargo.toml
amcelroy: The `pictorus-traits` crate is no_std and no alloc already. Do you mean to add in an `alloc` feature that is unused but stubbed out?

.github/workflows/ci-tests.yaml
amcelroy: The thinking was that we have core (`--no-default-features`), core + alloc (`--no-default-features --features alloc`), and core + alloc + std (`--all-features`). The CI checks `--all-targets --no-default-features --features alloc` are trying to ensure that core + alloc don't accidentally pull in std functionality or dependencies.

As a check I added `std::sync::Mutex` to a block in alloc_blocks and ran `--all-features` vs. `--no-default-features --features alloc`:
Screenshot 2026-05-28 at 12 38 29 PM

Dillonmcewan: Okay, that definitely makes sense for the `check` and/or `test` runs. Does it add any benefit to do that for `clippy`? Seems like clippy should just hit everything and deal with linting, and then check/test should deal with real errors like non-existent dependencies

.github/workflows/ci-tests.yaml
Dillonmcewan: As noted in other thread seems like this one is probably useful, but probably don't need both this and a separate lint

pictorus-blocks/Cargo.toml
Dillonmcewan: Oh, whoops. I mistook the traits.rs module changes as changes to the traits crate. Ignore this

Dillonmcewan (approved):
LGTM

PR #56, opened 2026-05-28
LSM Rust Code Prototype

jscatena88:
Alright here is where I ended up so far. 

The examples show what we would have to codegen, Although I think there are a few areas that a macro could clearly clean some of that up. I just didn't want to start chasing those down until we vetted the approach.

lib.rs is the real meat and potatoes of everything in terms of library/api/traits.

Happy to discuss tomorrow. I'll also probably try to add some more to this description tomorrow, I just wanted to get something up before logging off for the day


state-machine/examples/composite.rs
```diff
@@ -0,0 +1,329 @@
...
+    }
+}
+
+// ---- Flat input / output bundles ---------------------------------------
+
+#[derive(Debug, Default, Clone, Copy)]
+pub struct Input {
+    pub foo: Option<FooTransitions>,
+    pub bar: Option<BarTransitions>,
+    pub baz: Option<BazTransitions>,
```
Dillonmcewan: The example assumes that all transitions are unique, but one thing we'd like to support is shared events/transitions that can work at multiple levels of hierarchy. I.e. you might pass an `abort` event that different levels of the SM can respond to. Might not be a blocker to get started, but maybe worth considering how this might work

state-machine/examples/composite.rs
```diff
@@ -0,0 +1,329 @@
...
+    fn invalid_transition_does_not_change_state() {
+        let mut root = build_root();
+        let out = root.step(&Input {
+            foo: Some(FooTransitions::ToF1), // already at F1
+            ..Default::default()
+        });
+        assert_eq!(out.foo.state, FooStates::F1);
+        assert_eq!(out.foo.last_transition, None);
+    }
+}
```
Dillonmcewan: It doesn't look like any of the tests here cover the proposed priority-based step mechanism. Was that covered in the prototype impl?

state-machine/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+
+    fn reset(&mut self) {
+        let (ref mut n1, ref mut n2, ref mut n3) = *self;
+        n1.reset();
+        n2.reset();
+        n3.reset();
+    }
+}
+
+// Etc for (N1, N2, N3, N4) and so on... could be generated with a macro
```
Dillonmcewan: What do these variants represent? Parallel regions at a given level?

Dillonmcewan (commented):
Looks great overall. There were a couple cases I didn't see covered here

state-machine/src/lib.rs
jscatena88: Yep! It is easiest to see in the `parallel_children` example:
```
state_machine::children! {
    pub enum PowerChildren {
        On => (DisplayNode, AudioNode),
    }
}

pub type PowerNode = Node<PowerSpec, PowerChildren>;

pub fn build_power_node() -> PowerNode {
    let children = enum_map! {
        PowerStates::Off => PowerChildren::None,
        PowerStates::On => PowerChildren::On((build_display_node(), build_audio_node())),
    };
    PowerNode::new(
        children,
        |input: &Input| input.power,
        |output: &mut Output, snapshot| output.power = snapshot,
    )
}
```

This is taking advantage of `(DisplayNode, AudioNode)` implementing `NodeInterface` automatically since both elements of it do

state-machine/examples/composite.rs
jscatena88: Great reminder, I forgot to bring that up in the description. But as I was working through it I wasn't sure or at least didn't remember exactly how we intended that to work with the diagram interface we decided on. If every state machine only has one input being fed by one signal we can never end up with two requested transitions to have to use priority to decide between. 

I think the only way it could come into play is if we offered some way to have add additional transition inputs for a single statemachine, or if we had something analogous to the StateMachine handles of V2, where you could drop some sort of mutable reference to a statemachine in some other place in the block diagram

state-machine/examples/composite.rs
jscatena88: Yeah, this is not currently represented in any way. With the current scheme of coercing things in and out of f64s at the edges this could maybe imply making sure that the "same transition" has the same underlying repr value. Although this could get messy with priorities, if you wanted `Abort` to be the highest priority (e.g. 0 ) in one state machine and say the second highest in another state machine.

If you were willing to create some sort of "global priority" then you could potentially try to flatten all transitions into a single enum, although that could get tricky to make work with the `StateMachineSpec` pattern I am using. Forcing a more globally aware state machine specification system.

Another option could be some sort of translation mechanism that is able to map one set of transitions into another. As I write this, that could maybe intersect nicely with the "global transition list" I mentioned above. Maybe that gets put together like the flattened global inputs and outputs and each node does a `MyTransitionEnum::from(GlobalTransitionEnum)`?

I'm definitely not sure on the best path here

Dillonmcewan: Yeah it sounds like the options here would basically be:
1. Enums can somehow be shared between different levels of the SM
2. There are individual top-level inputs that get treated as distinct values that can get passed down the SM tree

I'm not 100% sure which direction we'll go, but maybe worth at least sketching out some ideas for both options until we can finalize the design

state-machine/examples/composite.rs
Dillonmcewan: Yeah it seems like this ties in closely with my other comment. If we go with the "each input is an enum" approach, then this is no longer relevant. But if we support top-level named inputs and/or guards that are not tied to transitions, then we will need a priority mechanism. Another one that might be worth sketching out even if we don't have the solution 100% sorted out.

jscatena88: Sounds good, I'll take a pass at what floating named transition requests to the top level looks like

state-machine/examples/transition_inputs.rs
```diff
@@ -0,0 +1,238 @@
...
+
+// ---- Flat input / output bundles --------------------------------------
+
+#[derive(Debug, Default, Clone, Copy)]
+pub struct Input {
+    pub power: Option<PowerTransitions>,
+    pub display: Option<DisplayTransitions>,
+    pub audio: Option<AudioTransitions>,
+    pub toggle: bool,
+    pub play: bool,
```
jscatena88: Add extra bool fields in the codegen-ed input struct

state-machine/examples/transition_inputs.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+pub type AudioNode = LeafNode<AudioSpec, Input, Output>;
+
+pub fn build_audio_node() -> AudioNode {
+    LeafNode::new_leaf(
+        |input: &Input| {
+            input
+                .audio
+                .prioritize_val(input.play, AudioTransitions::Play)
```
jscatena88: The trait I added gives us this `prioritize_val()` function that takes a bool and a variant. If the bool is None, it does nothing. If it is true it does essentially a [`reduce()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.reduce) on the `input.audio: Option<AudioTransitions>` and `Some( VAR)` where `VAR` is the second argument. It uses the normal `PartialOrd` less than operator as a comparison selecting the smaller of the two. This is the same prioritization we use in V2, and by default if you derive `PartialOrd` selects for the earliest variant in the enum declaration.

These can be arbitrarily chained so you could have something like
```
|input: &Input| {
    input
        .audio
        .prioritize_val(input.play, AudioTransitions::Play)
        .prioritize_val(input.pause, AudioTransitions::Pause)
       // etc....
}
```

This would also allow us to pretty easily handle having multiple state machines respond to the same named transition (i.e. bool input) with their own independent variant with that name (with whatever prioritization it was given locally)

state-machine/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+                    Self::None => {}
+                    Self::$first_variant(n) => n.reset(),
+                    $( Self::$variant(n) => n.reset(), )*
+                }
+            }
+        }
+    };
+}
+
+pub trait TransitionPrioritize {
```
jscatena88: This is how I get the chaining of the `prioritize_val()` functions

state-machine/examples/composite.rs
jscatena88: Just pushed up a commit and add comments to explain the changes. But honestly I'm super pleased with how easily and cleanly having name transition inputs was. Should be fairly easy to codegen and would support edges mapping to one SM or many.

It also by necessity introduces a concept of priority

state-machine/src/lib.rs
```diff
@@ -0,0 +1,838 @@
...
+        }
+    }
+}
+
+/// The provided default sink implementation: counts of emitted events, plus an optional
+/// log of their order that is enabled via the `event-log` feature (and requires `alloc`).
+pub struct Events<E: EnumArray<u32> + Copy> {
+    pub counts: EnumMap<E, u32>,
+    #[cfg(feature = "event-log")]
+    pub order: alloc::vec::Vec<E>,
```
amcelroy, line 32: If we know all of the events going into codegen, can we pre-allocate a heapless vector? Is this vector ever cleared?

state-machine/src/lib.rs
```diff
@@ -0,0 +1,838 @@
+#[cfg(feature = "event-log")]
+extern crate alloc;
```
amcelroy, line 7: Do we need a `#![no_std]` declaration?

state-machine/src/lib.rs
```diff
@@ -0,0 +1,895 @@
...
+        #[cfg(feature = "event-log")]
+        self.order.push(e);
+    }
+}
+
+/// A Guard is a function that takes the machine's InputData and returns a boolean indicating whether a transition is allowed to occur
+pub type Guard<D> = fn(&D) -> bool;
+/// A composite type representing a transition edge, consisting of an optional guard, an optional output event, and an optional target state
+/// where None indicates an "internal transition"
+pub type GuardedEdge<S, D, O> = (Option<Guard<D>>, Option<O>, Option<S>);
```
amcelroy: Nit: can the tuple order match the generic order?

state-machine/src/lib.rs
```diff
@@ -0,0 +1,895 @@
...
+            children: enum_map! {
+                Top::Active  => TopChildren::ActiveKids((Node::new_leaf(),  Node::new_leaf())),
+                Top::Standby => TopChildren::StandbyKids((Node::new_leaf(), Node::new_leaf())),
+            },
+        }
+    }
+
+    // ── Order-preserving sink (no alloc) ─────────────────────────────────────
+    struct RecordingSink<O: Copy> {
+        log: [Option<O>; 16],
```
amcelroy, line 908: So we could refactor for a heapless approach on Sinks?

state-machine/src/lib.rs
```diff
@@ -0,0 +1,895 @@
...
+        StandbyKids((Leaf<PowerSpec>, Leaf<NetworkSpec>)), // Power ∥ Network
+    }
+
+    impl NodeInterface for TopChildren {
+        type InputEvent = Ev;
+        type InputData = Data;
+        type OutputEvent = Out;
+        fn select(&mut self, input_event: Self::InputEvent, input_data: &Self::InputData) -> bool {
+            match self {
+                TopChildren::ActiveKids(t) => t.select(input_event, input_data),
```
amcelroy, line 863: The NodeInterface section of code seems a little mysterious when I try to frame it with the spec.
- Does this need to be codegened at each layer? 
- Will the user potentially be debugging `NodeInterface`?

amcelroy (commented):
Great job, this is a really nice first pass and implementation of the spec. I don't see any red flags, other than that it is a pretty dense set of functionality, like CoreLib. The requirement for arbitrary, nested, re-usable, and statically allocated SM means everything has to be encoded up front, so a lot of the complexity is chasing the generics through.

- Would most of the codegen look like the setup code in the unit test section? 
- You mentioned an example coming soon, a crate level README might be nice as well that summarizes the spec. Maybe walking through the example w/ the SM image?

state-machine/src/lib.rs
jscatena88: Unfortunately the `order` would be a little hard to size appropriately since it encodes every event that fires which could add up to much more elements than the number of possible events. (You could imagine a single input event could enter a state that has a super deep nest of children that spins up a massive number of OnEnter events and a similar number of OnExit events when it exits that state and this could happen several times in a single execution).

This alloc `order` is mainly to support debugging on std/sim. On the other hand since it is just an implementer of the `EventSink` trait there is no reason we couldn't potentially make an implementation that say dumps the output stream to a debug interface on an embedded target.

If I'm honest I haven't thought through the exact semantics of its lifecycle in a broader app, I just put something there for now

state-machine/src/lib.rs
jscatena88: Yep! I was mostly focused on the overarching architecture so there are definitely house keeping tasks to be done before we make it a real `pictorus-` crate or fold it into one of the existing crates

state-machine/src/lib.rs
jscatena88: There is certainly no reason you couldn't use heapless, This was just a super quick testing mock, but with it being based around the `EventSink` trait we can have any number of underlying implementations for outputing to RTT, a Heapless Vec, and alloc Vec, etc.

state-machine/src/lib.rs
jscatena88: Having a children enum that impls `NodeInterface` is the mechanism that allows hierarchical nesting.

 In the last version, to the old spec, I used a macro `children!()` to automatically handle most of this. I'm thinking I'll do the same again this time. It is more or less just a way to deal with having disparate types for all the states potential children and then dispatching function calls to the appropriate children when a state is active. So every implementation is pretty mechanical to crank out.

But hopefully a user will never have to think about these internals

state-machine/examples/audio_player.rs
```diff
@@ -0,0 +1,729 @@
...
+    type OutputEvent = OutputEvent;
+
+    const EDGES: EdgeTable<Player, InputEvent, InputData, OutputEvent> = &[
+        (
+            Player::Active,
+            &[
+                (
+                    InputEvent::Sleep,
+                    &[(
+                        Some(|c: &InputData| c.battery > 0),
```
Dillonmcewan: Nit: These transition defs might be more readable with structs/enums. This seems to work just fine as-is, but hard to tell what's going on without additional context

state-machine/examples/audio_player.rs
```diff
@@ -0,0 +1,729 @@
...
+// wins; the final unguarded edge is an *internal* transition (target = None)
+// that acts as the catch-all "nothing special happened this tick".
+struct PlayerSpec;
+impl StateMachineSpec for PlayerSpec {
+    type State = Player;
+    type InputEvent = InputEvent;
+    type InputData = InputData;
+    type OutputEvent = OutputEvent;
+
+    const EDGES: EdgeTable<Player, InputEvent, InputData, OutputEvent> = &[
```
Dillonmcewan: Nit: Would lean towards calling things "Transition" rather than "Edge" for state machines

state-machine/examples/audio_player.rs
```diff
@@ -0,0 +1,729 @@
...
+        }
+    }
+    fn on_exit(s: Player) -> Option<OutputEvent> {
+        match s {
+            Player::Active => Some(OutputEvent::ActiveExit),
+            Player::Standby => Some(OutputEvent::StandbyExit),
+            Player::Off => None,
+        }
+    }
+    fn during(s: Player) -> Option<OutputEvent> {
```
Dillonmcewan, line 235: Might be worth circling up on this one again. This is a pretty distinct departure from UML state machines. Want to clarify the behavior/use case before we commit to this one

state-machine/examples/audio_player.rs
```diff
@@ -0,0 +1,729 @@
...
+// heterogeneous subtrees.
+type Leaf<S> = LeafNode<S, InputEvent, InputData, OutputEvent>;
+
+enum PlayerChildren {
+    Active((Leaf<AudioSpec>, Leaf<ScreenSpec>)), // Audio ∥ Screen
+    Standby((Leaf<PowerSpec>, Leaf<NetSpec>)),   // Power ∥ Net
+    Off(NoChildren<InputEvent, InputData, OutputEvent>), // simple state: no children
+}
+
+impl NodeInterface for PlayerChildren {
```
Dillonmcewan: Another naming nit here. Not obvious to me what `Node` means in the context of an SM. Should this be called `StateInterface`? Or `Vertex` (if we want to go with UML naming)? Might be worth a pass at naming things to try to align with our UI and/or UML

state-machine/src/lib.rs
jscatena88: Ok examples are up, and I added the `children!` macro. Let me know how that reads for you

state-machine/examples/audio_player.rs
Dillonmcewan: Alternatively could just come up with a macro that's easier to read

state-machine/src/lib.rs
```diff
@@ -0,0 +1,1144 @@
...
+        #[cfg(feature = "event-log")]
+        self.order.push(e);
+    }
+}
+
+/// A Guard is a function that takes the machine's InputData and returns a boolean indicating whether a transition is allowed to occur
+pub type Guard<D> = fn(&D) -> bool;
+/// A composite type representing a transition edge, consisting of an optional guard, an optional output event, and an optional target state
+/// where None indicates an "internal transition"
+pub type GuardedEdge<D, O, S> = (Option<Guard<D>>, Option<O>, Option<S>);
```
Dillonmcewan: I would just call this a `Transition` I think. Or a `TransitionDef`

state-machine/src/lib.rs
```diff
@@ -0,0 +1,1144 @@
...
+    fn emit(&mut self, e: E) {
+        self.counts[e] += 1;
+        #[cfg(feature = "event-log")]
+        self.order.push(e);
+    }
+}
+
+/// A Guard is a function that takes the machine's InputData and returns a boolean indicating whether a transition is allowed to occur
+pub type Guard<D> = fn(&D) -> bool;
+/// A composite type representing a transition edge, consisting of an optional guard, an optional output event, and an optional target state
```
Dillonmcewan: Another naming nit "Output event" --> "Transition action"

state-machine/src/lib.rs
```diff
@@ -0,0 +1,1144 @@
...
+    fn default() -> Self {
+        Self {
+            current: S::State::default(),
+            pending: None,
+        }
+    }
+}
+
+/// Defines the interface for a composable element of a hierarchical/parallel state machine tree
+pub trait NodeInterface {
```
Dillonmcewan: I think this would be a "composite state" and the parallel things would be called "regions"

state-machine/src/lib.rs
```diff
@@ -0,0 +1,1144 @@
...
+                    $( Self::$variant(c) => c.reset(), )*
+                }
+            }
+        }
+    };
+}
+
+/// The primary runtime interface for users of this library: a `StateMachineRoot` is a wrapper around the top-level
+/// `Node` that provides a simple API for stepping the machine with input events and data, and for creating the machine with an initial node and sink.
+pub struct StateMachineRoot<N: NodeInterface> {
```
Dillonmcewan: I would expect the top-level thing to be called a "StateMachine". Everything under that is a "State". If it has children, it's a "Composite State", otherwise it's a "Simple State"

Dillonmcewan (commented):
In terms of logic/functionality this looks totally correct. I think biggest piece of feedback is that we should try to align the naming on our UI/UML. It is pretty hard to piece together what's happening in relation to the graphic SM def when everything uses different names. Also probably some opportunities to clean things up with macros. Both easy fixes though, so overall looks great!

state-machine/examples/audio_player.rs
```diff
@@ -0,0 +1,737 @@
...
+    > = TransitionTable::new(&[
+        StateTransitions {
+            source: Player::Active,
+            events: &[
+                EventTransitions {
+                    event: InputEvent::Sleep,
+                    transitions: &[Transition {
+                        guard: Some(|c: &InputData| c.battery > 0),
+                        action: Some(OutputEvent::GoStandby),
+                        destination: Some(Player::Standby),
```
Dillonmcewan: Nit: I think `target` is typically used here

state-machine/examples/audio_player.rs
```diff
@@ -0,0 +1,737 @@
...
+                    }],
+                },
+                EventTransitions {
+                    event: InputEvent::Fault,
+                    transitions: &[Transition {
+                        guard: None,
```
Dillonmcewan, line 156: I think it's moderately more readable with the structs here vs the old tuple approach. Not married to either one, so will defer to your preference

Dillonmcewan (approved):
LGTM. Much easier to understand with the new names. Only had one minor nit on use of `destination` vs. `target`

state-machine/examples/audio_player.rs
jscatena88: Yeah I think it is a little easier to read this way even if it is very verbose. But I think for our codegen usecase where a user will probably just need to read these once in a while and never really writing them this is probably the right trade off.

PR #60, opened 2026-06-22
Corelib processblock wrapper

jscatena88:


amcelroy (commented):
Nice job, looks like a pretty clean wrapping.

- Should there be unit tests for this this like other blocks or are we comfortable with test coverage elsewhere? 
- Should this be in pictorus-blocks? Guess we can always swap it over later; we also have Input/Output blocks implemented in platform specific crates, so there is precedence for keeping it here. Probably also keeps dependencies a little cleaner as well. 
- Not sure if we've discussed this elsewhere, but there is a real possibility LSMs will have more than 8 inputs and outputs with all the events and input data, it doesn't look like anything in this code would hinder that.

jscatena88:
There should 100% be some unit tests, I'll add those.

Yeah I'm still unsure where any of this should live. I'm gonna leave everything where it is for now to sort out later unless anyone has strong feelings

Yep, I definitely wrote this with an eye towards being able to use the array output trick if need be, I'll make sure both that and tuple output are covered in tests

state-machine/src/corelib_wrapper.rs
Dillonmcewan, line 1: Nit: Might call this module `block_wrapper` or `block_adapter` instead

Dillonmcewan: Also could use a docstring on this module

Dillonmcewan (approved):
LGTM with a couple optional comments

PR #61, opened 2026-06-23
Implement Orphan Rule fix for Process Block Adapter

jscatena88:
My original implementation ended up getting caught up in the Orphan Rule. For some reason I thought that because the underlying `StateDiagardam<SDS: StateDiagramSpec>` would be spared since the generic is local to the crate doing the impl, but that was not the case

### DIFF
--- a/state-machine/src/process_block_adapter.rs
+++ b/state-machine/src/process_block_adapter.rs
@@ -1,29 +1,56 @@
 //! This module provides a struct and set of traits that allow state machines to be used as a [`pictorus_traits::ProcessBlock`] in a Pictorus diagram.
 //! It is a bridge between the generic state machine abstractions and the Pictorus diagram abstractions.
 //!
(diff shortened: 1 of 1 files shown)


state-machine/src/process_block_adapter.rs
amcelroy: A comment here on the structure of the input (1 input data, 2 events, data is first, events second) might be good; it wasn't too bad to get via debugging the example and going through the `FloatSignalConverter`. On the block diagram, we have events first, then data beneath it, so it was a little reverse from what was mapped in my head.

amcelroy (approved):
LGTM, seems like a good solution to map arbitrary input array into the data and signals used to drive the LSM.

state-machine/src/process_block_adapter.rs
jscatena88: Ahh great call out, I was able to move that around pretty trivially to match expectations

PR #62, opened 2026-07-02
Initialization now occurs on the first execution instead of construction

jscatena88:


state-machine/src/process_block_adapter.rs
```diff
@@ -67,12 +67,40 @@ impl Parameter {
...
+    Initialized(StateMachine<SD>),
+    #[default]
+    None,
+}
+
+impl<SD: StateDiagramInterface> StateMachineStorage<SD>
+where
+    SD::OutputEvent: EnumArray<u32> + Copy,
+{
+    fn get_initialized(&mut self, events: &mut Events<SD::OutputEvent>) -> &mut StateMachine<SD> {
```
amcelroy, line 89: Maybe some doc strings about the need for this and some notes on the match arms and reason for `take`?

Dillonmcewan (approved):

amcelroy (approved):
LGTM, nit about adding some extra doc strings.

state-machine/src/process_block_adapter.rs
jscatena88: Good idea! It is definitely some proper borrow checker shenanigans. I just added some comments to help explain

PR #63, opened 2026-07-02
Patch custom fork of fmu-runner until insecure dep is upgraded

jscatena88:
I made a fork that upgrades to the secure version of the dep. I have a PR open with the project to get my upgrade merged in. We can remove this once that happens

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,3 +25,6 @@ homepage = "https://www.pictor.us/"
 repository = "https://github.com/Pictorus-Labs/pictorus-rs"
 keywords = ["embedded", "modeling", "simulation", "fmi"]
 categories = ["embedded", "hardware-support", "no-std"]
+
+[patch.crates-io]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -25,3 +25,6 @@ homepage = "https://www.pictor.us/"
 repository = "https://github.com/Pictorus-Labs/pictorus-rs"
 keywords = ["embedded", "modeling", "simulation", "fmi"]
 categories = ["embedded", "hardware-support", "no-std"]
+
+[patch.crates-io]
+fmu-runner = { git = "https://github.com/jscatena88/fmu-runner-rs", branch = "main" }
```
amcelroy: Should this be a fork owned by Pictorus-Labs? Probably doesn't matter for now.

amcelroy (approved):
LGTM

Cargo.toml
jscatena88: Yeah that makes sense, just changed it to use a pictorus-labs fork of fmu-runner

PR #64, opened 2026-07-07
Make crates more platform independent where possible

jscatena88:
This PR fixes csv_logger to work on Windows as well as Linux. It renames `LinuxLogger` to be `StdLogger`. And finally it breaks out `pictorus-linux`, creating another `pictorus-std` crate to hold any functionality that relies on having `std` but does not necessarily require linux


pictorus-internal/src/loggers/csv_logger.rs
```diff
@@ -23,19 +24,20 @@ pub struct CsvLogger {
...
 impl CsvLogger {
     pub fn new(csv_log_period: Duration, output_path: std::path::PathBuf) -> Self {
-        let mut file_obj = File::create("/dev/null").unwrap();
-        if !csv_log_period.is_zero() {
+        let writer = if !csv_log_period.is_zero() {
             info!("DataLogger CSV output period: {csv_log_period:?}");
             info!("Streaming data output to file: {}", output_path.display());
-            file_obj = File::create(std::path::PathBuf::from(&output_path)).unwrap();
+            let file_obj = File::create(&output_path).unwrap();
+            Box::new(BufWriter::with_capacity(65536, file_obj)) as Box<dyn Write>
```
amcelroy, line 31: Was the upgrade to Box so that we could swap out `Write` impl more easily instead of forced into a `BufWriter`?

amcelroy (approved):
LGTM! Nice refactor moving clock, delay, serial, and UDP to common STD crate for Windows builds.

pictorus-internal/src/loggers/csv_logger.rs
jscatena88: Yep exactly. It lets us use a BufWriter over a file if we need it and the std::io::sink if we aren't writing to a file (instead of /dev/null  which is Linux specific)

PR #65, opened 2026-07-22
Rework Renesas GPIO to match reworked HAL

amcelroy:
The HAL was reworked to allow GPIO for all valid ports / pins. The rework changed the generic to an OutputPin and InputPin instead of a pin number.

Changes:
- Reworked Renesas RA4M2 InputBlock and OutputBlock to work with the updated HAL.
- Updated HAL crate rev in Cargo.toml


jscatena88 (approved):
Looks good! Sorry I thought I had already approved this yesterday when I looked at it

PR #66, opened 2026-07-23
Add support for f32 default type

jscatena88:
This PR adds `f32` for every block that was `f64` only except for blocks where it doesn't apply. This is based on past investigation of the blocks.

The only remaining blocks without `f32` support should be the following:  BytesLiteral, I2cInput, I2cOutput, SerialReceive, SpiReceive, UdpReceive, UdpTransmit, and Fmu. These all don't logically make sense to mess with since they either deal only with ByteSliceSignal signals, or for FMU `f64` is the only officially supported data type.

### DIFF
--- a/pictorus-blocks/src/alloc_blocks/bytes_unpack_block.rs
+++ b/pictorus-blocks/src/alloc_blocks/bytes_unpack_block.rs
@@ -5,6 +5,7 @@ use crate::{
     traits::Float,
 };
 use generic_array::{ArrayLength, GenericArray};
+use num_traits::AsPrimitive;
 use typenum::{Const, NonZero, Sub1, ToUInt, B1, U};
 
 use crate::byte_data::{parse_byte_data_spec, try_unpack_data, ByteOrderSpec, DataType};
@@ -107,42 +108,31 @@ pub trait Unpack: Float {
         -> (Option<Self>, &[u8]);
 }
 
-impl Unpack for f64 {
+impl<F> Unpack for F
+where
+    F: Float,
+    u8: AsPrimitive<F>,
+    i8: AsPrimitive<F>,
+    u16: AsPrimitive<F>,
+    i16: AsPrimitive<F>,
+    u32: AsPrimitive<F>,
+    i32: AsPrimitive<F>,
+    u64: AsPrimitive<F>,
+    i64: AsPrimitive<F>,
+    u128: AsPrimitive<F>,
+    i128: AsPrimitive<F>,
+    f32: AsPrimitive<F>,
+    f64: AsPrimitive<F>,
+{
     fn unpack(
         data: &[u8],
         data_type: DataType,
         byte_order: ByteOrderSpec,
     ) -> (Option<Self>, &[u8]) {
         let val = match byte_order {
-            ByteOrderSpec::BigEndian => {
-                try_unpack_data::<f64, byteorder::BigEndian>(data, data_type)
-            }
+            ByteOrderSpec::BigEndian => try_unpack_data::<F, byteorder::BigEndian>(data, data_type),
             ByteOrderSpec::LittleEndian => {
-                try_unpack_data::<f64, byteorder::LittleEndian>(data, data_type)
-            }
-        }
-        .ok();
-        let advanced_data = if val.is_some() {
-            &data[data_type.byte_size()..]
-        } else {
-            data
-        };
-        (val, advanced_data)
-    }
-}
-
-impl Unpack for f32 {
-    fn unpack(
-        data: &[u8],
-        data_type: DataType,
-        byte_order: ByteOrderSpec,
-    ) -> (Option<Self>, &[u8]) {
-        let val = match byte_order {
-            ByteOrderSpec::BigEndian => {
-                try_unpack_data::<f32, byteorder::BigEndian>(data, data_type)
-            }
-            ByteOrderSpec::LittleEndian => {
-                try_unpack_data::<f32, byteorder::LittleEndian>(data, data_type)
+                try_unpack_data::<F, byteorder::LittleEndian>(data, data_type)
             }
         }
         .ok();
--- a/pictorus-blocks/src/alloc_blocks/bytes_unpack_block.rs
+++ b/pictorus-blocks/src/alloc_blocks/bytes_unpack_block.rs
@@ -2,7 +2,7 @@ use core::time::Duration;
 
 use crate::{
     stale_tracker::{duration_from_ms_f64, StaleTracker},
-    traits::Scalar,
+    traits::Float,
 };
 use generic_array::{ArrayLength, GenericArray};
 use typenum::{Const, NonZero, Sub1, ToUInt, B1, U};
@@ -12,17 +12,17 @@ use pictorus_traits::{ByteSliceSignal, PassBy, ProcessBlock};
 
 /// Unpacks a byte slice into a specified number of outputs based on the provided data types and byte order.
 ///
-/// The output is `[f64; N]` where the last element is a `1.0`/`0.0` validity flag — true while
-/// the most recent successful unpack is still within the configured `stale_age`.
-pub struct BytesUnpackBlock<const N: usize> {
-    buffer: [f64; N],
+/// The output is `[O; N]` (`O` defaults for `f64` unless otherwise set) where the last element is a `1.0`/`0.0`
+/// validity flag — true while the most recent successful unpack is still within the configured `stale_age`.
+pub struct BytesUnpackBlock<const N: usize, O: Unpack = f64> {
+    buffer: [O; N],
     stale_check: StaleTracker,
 }
 
-impl<const N: usize> Default for BytesUnpackBlock<N> {
+impl<const N: usize, O: Unpack> Default for BytesUnpackBlock<N, O> {
     fn default() -> Self {
         BytesUnpackBlock {
-            buffer: [0.0; N],
+            buffer: [O::zero(); N],
             stale_check: StaleTracker::default(),
         }
     }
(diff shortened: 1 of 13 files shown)


pictorus-blocks/src/core_blocks/change_detection_block.rs
```diff
@@ -15,14 +16,14 @@ use strum::EnumString;
 ///
 /// The first execution of this block has no value to compare with
 /// so it will always emit `false`
-pub struct ChangeDetectionBlock<T: Apply> {
+pub struct ChangeDetectionBlock<T: Apply<O>, O: Scalar = f64> {
```
jscatena88, line 19: I used this trick a few times, it should make it so that the changes here don't break codegen. But I think it could be worth considering removing these defaults once codegen is update.

pictorus-blocks/src/core_blocks/counter_block.rs
```diff
@@ -50,59 +52,69 @@ impl<T: Apply> ProcessBlock for CounterBlock<T> {
...
 
-pub trait Apply: Pass {
+pub trait Apply<O>: Pass {
     type Counter: Default + Pass;
     fn apply<'a>(count: &'a mut Self::Counter, input: PassBy<Self>) -> PassBy<'a, Self::Counter>;
 }
 
-impl<I: Scalar, R: Scalar> Apply for (I, R) {
-    type Counter = f64;
+impl<I: Scalar, O: Scalar + num_traits::Zero + num_traits::One, R: Scalar> Apply<O> for (I, R) {
```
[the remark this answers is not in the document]
jscatena88: This is implied by the definition of the `Zero` trait. In the `num_traits` crate it is defined as:
```
pub trait Zero: Sized + Add<Self, Output = Self> { ... }
```

pictorus-blocks/src/alloc_blocks/bytes_unpack_block.rs
```diff
@@ -114,9 +114,35 @@ impl Unpack for f64 {
...
+        let advanced_data = if val.is_some() {
+            &data[data_type.byte_size()..]
+        } else {
+            data
+        };
+        (val, advanced_data)
+    }
+}
+
+impl Unpack for f32 {
```
amcelroy: It seems like there is duplicate logic between f32/f64 impl for both `bytes_pack_block` and `bytes_unpack_block`. Is there a way to consolidate the logic with macros or a generic?

pictorus-blocks/src/byte_data.rs
```diff
@@ -194,60 +194,92 @@ pub fn parse_string_to_read_delimiter(data: &str) -> (Vec<u8>, Vec<usize>) {
...
-        DataType::I48 => Endian::read_i48(buf) as f64,
-        DataType::U64 => Endian::read_u64(buf) as f64,
-        DataType::I64 => Endian::read_i64(buf) as f64,
-        DataType::U128 => Endian::read_u128(buf) as f64,
-        DataType::I128 => Endian::read_i128(buf) as f64,
-        DataType::F32 => Endian::read_f32(buf) as f64,
-        DataType::F64 => Endian::read_f64(buf),
+        DataType::U8 => buf[0].as_(),
+        DataType::I8 => AsPrimitive::<i8>::as_(buf[0]).as_(),
+        DataType::U16 => Endian::read_u16(buf).as_(),
```
amcelroy, line 225: `.as_()` is super interesting nomenclature, had to read through [AsPrimitive](https://docs.rs/num-traits/latest/num_traits/cast/trait.AsPrimitive.html), very cool!

pictorus-blocks/src/traits.rs
```diff
@@ -25,51 +25,121 @@ pub trait Scalar:
     /// Returns true if the scalar is truthy
     /// Truthiness is defined as not equal to zero
     fn is_truthy(&self) -> bool;
+
+    fn from_bool(value: bool) -> Self;
```
amcelroy, line 29: Great idea

pictorus-blocks/src/byte_data.rs
```diff
@@ -382,13 +414,25 @@ mod tests {
...
+        try_pack_data::<f64, BigEndian>(&mut packed_data, input, dt).unwrap();
+        let unpacked_data = try_unpack_data::<f64, BigEndian>(&packed_data, dt).unwrap();
+
+        assert_eq!(input, unpacked_data);
+    }
+
+    #[test]
+    fn test_try_unpack_data_and_try_pack_data_f32() {
+        let input: f32 = 1234.5678;
+        let dt = DataType::F64;
```
amcelroy, line 431: Should this be F32?

pictorus-blocks/src/core_blocks/timer_block.rs
```diff
@@ -35,54 +37,57 @@ impl Parameters {
...
         &mut self,
         parameters: &Self::Parameters,
         context: &dyn pictorus_traits::Context,
         input: PassBy<Self::Inputs>,
     ) -> PassBy<'_, Self::Output> {
-        let time = context.time().as_secs_f64();
+        let time = O::from_duration(context.time());
 
-        let trigger_high = input > 0.0;
+        let trigger_high = input.is_truthy();
```
amcelroy, line 90: This might be a breaking change, as negative numbers will be a trigger, where before it was just positive numbers, though this seems more consistent with how we handle thruthy-ness.

pictorus-blocks/src/core_blocks/pid_block.rs
```diff
@@ -129,25 +130,13 @@ where
...
-
-impl<const ND_SAMPLES: usize, const NROWS: usize, const NCOLS: usize, R: Scalar> HasIc
-    for PidBlock<Matrix<NROWS, NCOLS, f64>, R, ND_SAMPLES>
+impl<T: ComponentOps, R: Scalar, const ND_SAMPLES: usize> HasIc for PidBlock<T, R, ND_SAMPLES>
+where
+    DerivativeBlock<T, ND_SAMPLES>:
+        ProcessBlock<Output = T, Inputs = T, Parameters = DerivativeParameters<T>> + HasIc,
+    IntegralBlock<(T, R)>:
+        ProcessBlock<Output = T, Inputs = (T, R), Parameters = IntegralParameters<(T, R)>> + HasIc,
+    (T, R): IntegralApply<Output = T, Float = T::Float> + for<'a> Pass<By<'a> = (PassBy<'a, T>, R)>,
```
amcelroy, line 142: This got a little spicy. Don't suppose there is a comment blurb we could add here for future explorers.

pictorus-blocks/src/std_blocks/system_time_block.rs
amcelroy, line 1: Is this one that we should consider as f64 only, at least for the Epoch return value? There may be some rough edges at f32.

https://stackoverflow.com/a/38594632/2521274

amcelroy (approved):
Nice job Jason; I enjoy learning from your Rust skills and picked up a couple of new lessons. Couple of comments, no blockers.

pictorus-blocks/src/core_blocks/timer_block.rs
Dillonmcewan: Yep. I would say the previous implementation was actually a bug based on our documented boolean handling. This seems more correct

pictorus-blocks/src/std_blocks/system_time_block.rs
jscatena88: Ahh, that is a great point. It looks like with 24 bits you can represent all integers <= 33,554,430 which is already larger than the unix epoch time right now of ~1,784,916,438. So this would necessarily induce precision loss. 

I can't think of a good way to disallow `f32` for only one value of the `SystemTimeEnum` since that is a runtime parameter.

The way I see it we have 3 options:
1. We could make the parameter a type state style parameter and that would allow us to do just that and not allow `f32` for specifically the Epoch case
2. When set to `f32` we could use `epoch_time % 2^24` which would allow apps that are using it as a sort of counter to function correctly with a discontinuity every 388 days (although an app running at just the wrong time would catch it, which isn't great)
3. Just not support `f32` for this block

2 seems like a non-starter, 1 is possible but adds complexity, it could be an eventual solutions, 3 seems like the right move for right now

Dillonmcewan: Yep, that's a good flag. That might be very difficult to represent in the sizing system if we only restrict it to the Epoch mode though. Maybe worth just flagging in docs that it's a really bad idea to use f32 for epoch time. Might be worth rethinking this block a bit once we support non-float values. I would typically lean toward using uints here where the precision level is configurable (i.e. EpochSec vs. EpochMs), but out of scope for this PR

Dillonmcewan: I think just requiring f64 for this one and discussing later is a good solution for now 👍

Dillonmcewan (approved):
LGTM!

pictorus-blocks/src/alloc_blocks/bytes_unpack_block.rs
jscatena88: Yep, good point I was able to make these both generic with the same bounds as the `try_pack_data` and `try_unpack_data` functions respectively

pictorus-blocks/src/core_blocks/pid_block.rs
jscatena88: Yep, I'll add something here

pictorus-blocks/src/byte_data.rs
jscatena88: That is just the type we are packing it into before unpacking it out of it. `f64` was chosen just because its not a no-op but also won't loose any precision like a trip through say a `u8` would

pictorus-blocks/src/std_blocks/system_time_block.rs
jscatena88: Ok, I just reverted this for now so it is back to only support `f64`

PR #68, opened 2026-08-11
Add Cast Block

amcelroy:
This PR adds Cast Block, which allows casting from booleans and numerics into other booleans and numerics.

Changes:
- Rework Scalar trait a bit. It requires implementors to also implement Into<f64>, which is a no go for u64 and i64. Replaced with a trait function `to_f64_lossy` which casts to f64 using `as`. Implemented this for numerics and bools.
- Rework CustomBlock to use `to_f64_lossy` instead of `Into<f64>`.
- Add `is_truthy` and `from_bool` for i64 and u64
- Add cast_block.rs which implements casting for Scalar and Matrix<_,_,Scalar>.

### DIFF
--- a/pictorus-blocks/src/core_blocks/cast_block.rs
+++ b/pictorus-blocks/src/core_blocks/cast_block.rs
@@ -1,0 +1,324 @@
+use crate::traits::Scalar;
+use pictorus_traits::{Context, Matrix, Pass, PassBy, ProcessBlock};
+
+/// Parameters for the Cast block
+///
+/// The destination type is carried in the block's generics rather than here, since it has
+/// to be known at compile time.
+#[doc(hidden)]
+pub struct Parameters;
+
+impl Default for Parameters {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Parameters {
+    pub fn new() -> Parameters {
+        Parameters {}
+    }
+}
+
+/// Converts a scalar element from one type to another.
+///
+/// This is deliberately not built on `num_traits::AsPrimitive` (as of 0.2.19), which has 
+//  incomplete bool conversion. Instead the three cases are handled explicitly:
+/// - converting to `bool` tests truthiness
+//  - converting from `bool` maps to 0/1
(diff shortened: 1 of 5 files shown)


pictorus-blocks/src/core_blocks/cast_block.rs
```diff
@@ -0,0 +1,324 @@
...
+    I: CastTo<O>,
+    O: Pass + Default,
+{
+    fn default() -> Self {
+        Self {
+            buffer: O::default(),
+            phantom: core::marker::PhantomData,
+        }
+    }
+}
```
jscatena88: nit: I believe you should be able to derive this `Default` impl since it is using the default forms of its members

pictorus-traits/src/lib.rs
```diff
@@ -419,34 +419,44 @@ where
...
-impl Sealed for u16 {}
-
-impl Scalar for i16 {}
-impl Sealed for i16 {}
-
-impl Scalar for u32 {}
-impl Sealed for u32 {}
+///
+/// Note that this trait deliberately doesn't require `Into<f64>`; the full range of
+/// `u64`/`i64` is not representable in an `f64`, so they don't work with `.into()`.
```
jscatena88: nit: probably unnecessary context

pictorus-blocks/src/core_blocks/cast_block.rs
```diff
@@ -0,0 +1,324 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Parameters {
+    pub fn new() -> Parameters {
+        Parameters {}
+    }
+}
```
jscatena88, line 16: Antoher nitty nit: I think you should be able to derive `Default` and then change the new impl to:
```rust
impl Parameters {
  pub fn new() -> Parameters {
    Self::default()
  }
}
```

jscatena88 (approved):
Looks good, a few nits on comments or conciseness of the code but it all looks correct

PR #69, opened 2026-08-17
Flush out blocks generics for the wider set of scalars where possible

jscatena88:
All blocks should be covered now. In a few cases this change exposes us to new opportunities to panic in cases of arithmetic overflow/underflow  (in debug builds it panics, in release builds it will wrap). I have annotated as many of these cases as I could identify with `#[should_fail]` unit tests so they are documented and  we can revisit them in the future

Of note the following were skipped or limited in their impls:
- System Time Block: the inability to represent unix time in most scalars means I left it alone
- ADC was skipped since I didn't think it made sense to have arbitrary casting in the block if the type retrieved from HW is always the same, it would probably be better to have people use a cast block. Worth noting that type from HW is not `f64`, it could make sense to go back at some point and make it output honest to what the HW outputs
- JSON Load: These come from json as `i64`/`u64`/`f64`, we are already doing a lossy conversion to `f64` for the interger cases, but by similar reasoning to ADC it figured it doesn't make sense to add conversion in the block that is essentially just a hidden cast block
- CanTX and CanRX: Again DBCs output types according to their call back ABI, we should look at these deeper at a later point
- Bytes Unpack: Again this feels like it requires some deeper effort/ thought. In an ideal world we would actually output the types taken out of the byte input. This however would require us going from `[f64; N]` output to a tuple based output so we could have mixed types. This would again limit us to how high our tuple imples of `Pass` go (8 outputs total counting is_valid). That limit is why we switched to the array impl in the first place

### DIFF
--- a/pictorus-blocks/src/alloc_blocks/bytes_pack_block.rs
+++ b/pictorus-blocks/src/alloc_blocks/bytes_pack_block.rs
@@ -1,5 +1,5 @@
 use crate::byte_data::{parse_byte_data_spec, try_pack_data, ByteOrderSpec, DataType};
-use crate::traits::{Float, Scalar};
+use crate::traits::Scalar;
 use alloc::vec::Vec;
 use num_traits::AsPrimitive;
 use pictorus_traits::{ByteSliceSignal, Pass, PassBy, ProcessBlock};
@@ -61,7 +61,7 @@ pub trait AppendBytes: Scalar {
 
 impl<F> AppendBytes for F
 where
-    F: Float
+    F: Scalar
         + AsPrimitive<u8>
         + AsPrimitive<i8>
         + AsPrimitive<u16>
@@ -253,6 +253,91 @@ mod tests {
         let expected = {
             let mut expected = Vec::new();
             expected.write_i8(inputs as i8).unwrap();
+            expected
+        };
+
+        let output = block.process(&params, &context, inputs);
+        assert_eq!(output, expected.as_slice());
+        assert_eq!(block.buffer(), expected.as_slice());
+    }
+
+    #[test]
+    fn test_bytes_pack_block_1_input_u8_exact() {
+        let context = StubContext::default();
+        let params = Parameters::new(&["U8:BigEndian"]);
+        let mut block = BytesPackBlock::<u8>::default();
+
+        let output = block.process(&params, &context, 255u8);
+        assert_eq!(output, [255u8].as_slice());
+        assert_eq!(block.buffer(), [255u8].as_slice());
+    }
+
+    #[test]
+    fn test_bytes_pack_block_1_input_i16_negative() {
+        let context = StubContext::default();
+        let params = Parameters::new(&["I16:LittleEndian"]);
+        let mut block = BytesPackBlock::<i16>::default();
+
+        let expected = {
+            let mut expected = Vec::new();
+            expected
+                .write_i16::<byteorder::LittleEndian>(-12345)
+                .unwrap();
+            expected
+        };
+
+        let output = block.process(&params, &context, -12345i16);
+        assert_eq!(output, expected.as_slice());
+        assert_eq!(block.buffer(), expected.as_slice());
+    }
+
+    #[test]
+    fn test_bytes_pack_block_1_input_i64_beyond_f64_precision() {
+        // An i64 input larger than 2^53 packs exactly; routing it through an f64
+        // signal (the only option before int support) would have rounded it.
+        let context = StubContext::default();
+        let params = Parameters::new(&["I64:BigEndian"]);
+        let mut block = BytesPackBlock::<i64>::default();
+        let value = (1i64 << 53) + 1;
+
+        let expected = {
+            let mut expected = Vec::new();
+            expected.write_i64::<byteorder::BigEndian>(value).unwrap();
+            expected
+        };
+
+        let output = block.process(&params, &context, value);
+        assert_eq!(output, expected.as_slice());
+        assert_eq!(block.buffer(), expected.as_slice());
+    }
+
+    #[test]
+    fn test_bytes_pack_block_int_truncating_spec() {
+        // An input wider than its pack spec truncates with native `as` cast semantics.
+        let context = StubContext::default();
+        let params = Parameters::new(&["U8:BigEndian"]);
+        let mut block = BytesPackBlock::<u32>::default();
+
+        let output = block.process(&params, &context, 300u32);
+        assert_eq!(output, [300u32 as u8].as_slice());
+    }
+
+    #[test]
+    fn test_bytes_pack_block_mixed_input_types() {
+        let context = StubContext::default();
+        let params = Parameters::new(&["U8:BigEndian", "F32:BigEndian", "I16:LittleEndian"]);
+        let mut block = BytesPackBlock::<(u8, f32, i16)>::default();
+        let inputs = (42u8, 3.5f32, -1000i16);
+
+        let expected = {
+            let mut expected = Vec::new();
+            expected.write_u8(inputs.0).unwrap();
+            expected
+                .write_f32::<byteorder::BigEndian>(inputs.1)
+                .unwrap();
+            expected
+                .write_i16::<byteorder::LittleEndian>(inputs.2)
+                .unwrap();
             expected
         };
 
--- a/pictorus-blocks/src/alloc_blocks/bytes_pack_block.rs
+++ b/pictorus-blocks/src/alloc_blocks/bytes_pack_block.rs
@@ -312,6 +312,26 @@ mod tests {
     }
 
     #[test]
+    fn test_bytes_pack_block_1_input_u64_beyond_f64_precision() {
+        // A u64 input larger than 2^53 (and above i64::MAX) packs exactly; routing it
+        // through an f64 signal (the only option before int support) would have rounded it.
+        let context = StubContext::default();
+        let params = Parameters::new(&["U64:BigEndian"]);
+        let mut block = BytesPackBlock::<u64>::default();
+        let value = u64::MAX - 1;
+
+        let expected = {
+            let mut expected = Vec::new();
+            expected.write_u64::<byteorder::BigEndian>(value).unwrap();
+            expected
+        };
+
+        let output = block.process(&params, &context, value);
+        assert_eq!(output, expected.as_slice());
+        assert_eq!(block.buffer(), expected.as_slice());
+    }
+
+    #[test]
     fn test_bytes_pack_block_int_truncating_spec() {
         // An input wider than its pack spec truncates with native `as` cast semantics.
         let context = StubContext::default();
--- a/pictorus-blocks/src/alloc_blocks/switch_block.rs
+++ b/pictorus-blocks/src/alloc_blocks/switch_block.rs
@@ -109,8 +109,8 @@ pub trait ApplyInto<C: Scalar, const N: usize>: Pass + DefaultStorage {
     );
 }
 
-impl<C: Scalar, const N: usize> ApplyInto<C, N> for C {
-    fn apply_into(condition: C, cases: &[C; N], inputs: &[PassBy<C>; N], dest: &mut C) {
+impl<C: Scalar, S: Scalar, const N: usize> ApplyInto<C, N> for S {
+    fn apply_into(condition: C, cases: &[C; N], inputs: &[PassBy<S>; N], dest: &mut S) {
         for (idx, case) in cases.iter().enumerate() {
             if condition == *case {
                 let res = inputs[idx];
@@ -123,14 +123,14 @@ impl<C: Scalar, const N: usize> ApplyInto<C, N> for C {
     }
 }
 
-impl<C: Scalar, const NROWS: usize, const NCOLS: usize, const N: usize> ApplyInto<C, N>
-    for Matrix<NROWS, NCOLS, C>
+impl<C: Scalar, S: Scalar, const NROWS: usize, const NCOLS: usize, const N: usize> ApplyInto<C, N>
+    for Matrix<NROWS, NCOLS, S>
 {
     fn apply_into(
         condition: C,
         cases: &[C; N],
-        inputs: &[PassBy<Matrix<NROWS, NCOLS, C>>; N],
-        dest: &mut Matrix<NROWS, NCOLS, C>,
+        inputs: &[PassBy<Matrix<NROWS, NCOLS, S>>; N],
+        dest: &mut Matrix<NROWS, NCOLS, S>,
     ) {
         for (idx, case) in cases.iter().enumerate() {
             if condition == *case {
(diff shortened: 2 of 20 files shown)


pictorus-blocks/src/alloc_blocks/bytes_pack_block.rs
```diff
@@ -61,7 +61,7 @@ pub trait AppendBytes: Scalar {
 
 impl<F> AppendBytes for F
 where
-    F: Float
+    F: Scalar
```
jscatena88, line 64: This was a vestigial requirement, so removing it automatically opened up BytesPack

pictorus-blocks/src/alloc_blocks/switch_block.rs
```diff
@@ -109,8 +109,8 @@ pub trait ApplyInto<C: Scalar, const N: usize>: Pass + DefaultStorage {
     );
 }
 
-impl<C: Scalar, const N: usize> ApplyInto<C, N> for C {
-    fn apply_into(condition: C, cases: &[C; N], inputs: &[PassBy<C>; N], dest: &mut C) {
+impl<C: Scalar, S: Scalar, const N: usize> ApplyInto<C, N> for S {
+    fn apply_into(condition: C, cases: &[C; N], inputs: &[PassBy<S>; N], dest: &mut S) {
```
jscatena88, line 113: This was just a subtle bug that was preventing the condition from having a distinct type from the other inputs

pictorus-blocks/src/core_blocks/abs_block.rs
```diff
@@ -178,7 +179,17 @@ mod tests {
...
+    #[should_panic]
+    fn overflow_quirk() {
+        // According to num_traits docs this should return `::MIN` for signed integers, but instead it appears to panic:
+        // https://docs.rs/num-traits/latest/num_traits/sign/trait.Signed.html#tymethod.abs
+        let mut block = AbsBlock::<i8>::default();
+        let context = StubContext::default();
+
+        let output = block.process(&Parameter::new(), &context, i8::MIN);
+        assert_eq!(output, -128i8);
+    }
```
jscatena88: This was a weird one, the observed behavior doesn't match the `num_traits` doc comment, I might open an issue with them

pictorus-blocks/src/core_blocks/bias_block.rs
```diff
@@ -1,113 +1,87 @@
...
 {
-    type Output = Promotion<B, f32>;
-
-    fn apply<'s>(
-        store: &'s mut Self::Output,
-        input: PassBy<Self>,
-        offset: B,
-    ) -> PassBy<'s, Self::Output> {
-        let output =
-            <B as Promote<f32>>::promote_left(offset) + <B as Promote<f32>>::promote_right(input);
```
jscatena88, line 88: I just removed all the promote stuff here. This makes it so that the parameter and the input must be of the same type, which I think is fine

pictorus-blocks/src/core_blocks/compare_to_value_block.rs
```diff
@@ -239,12 +247,59 @@ mod tests {
...
-    test_compare_to_value!(u16, u16);
-    test_compare_to_value!(i32, i32);
-    test_compare_to_value!(u32, u32);
-    test_compare_to_value!(f32, f32);
-    test_compare_to_value!(f64, f64);
+    test_compare_to_value!(i8, u8, i16, u16, i32, u32, i64, u64, f32, f64);
+
+    #[test]
+    #[allow(clippy::bool_assert_comparison)]
+    fn test_compare_to_value_bool() {
```
jscatena88, line 254: The idea of less than and greater than applied to `bool` was a little foreign to me so I ran through all the cases here. The lint is because it wants me to use `assert!(condition);` or `assert!(!condition);` I personally think the `assert_eq!` makes things more clear in this case though

pictorus-blocks/src/alloc_blocks/bytes_pack_block.rs
```diff
@@ -261,6 +261,91 @@ mod tests {
...
+        };
+
+        let output = block.process(&params, &context, -12345i16);
+        assert_eq!(output, expected.as_slice());
+        assert_eq!(block.buffer(), expected.as_slice());
+    }
+
+    #[test]
+    fn test_bytes_pack_block_1_input_i64_beyond_f64_precision() {
+        // An i64 input larger than 2^53 packs exactly; routing it through an f64
```
amcelroy, line 296: Do we want any precision tests for f32 or u64?

pictorus-blocks/src/alloc_blocks/bytes_pack_block.rs
amcelroy, line 1: Good test expansion!

pictorus-blocks/src/alloc_blocks/switch_block.rs
amcelroy: Good catch that only showed up with the introduction to types.

pictorus-blocks/src/core_blocks/abs_block.rs
amcelroy: Do i16, i32, i64 have the same issue for ::MIN? What about for ::MAX?

pictorus-blocks/src/core_blocks/aggregate_block.rs
```diff
@@ -237,19 +233,75 @@ mod tests {
...
+        let input = Matrix {
+            data: [[34, 127], [31, 4]],
+        };
+        let output = block.process(&PARAM_SUM, &context, &input);
+        assert_eq!(output, 196);
+    }
+
+    #[test]
+    #[should_panic]
+    fn overflow_in_add_panics() {
```
amcelroy, line 283: Not arguing that this behavior is right or wrong in the context of the PR. We've discussed Result or Option signals, but I wonder if there should some default behavior if things like a Sum panic. Like a user input default value,   using `saturating_add` instead, or introducing an `is_valid` (though that would lead to more conditional blocks).

pictorus-blocks/src/core_blocks/bias_block.rs
```diff
@@ -1,113 +1,87 @@
...
 
-    fn apply<'s>(
-        store: &'s mut Self::Output,
-        input: PassBy<Self>,
-        offset: B,
-    ) -> PassBy<'s, Self::Output> {
-        let output =
-            <B as Promote<f64>>::promote_left(offset) + <B as Promote<f64>>::promote_right(input);
+    fn apply<'s>(store: &'s mut Self, input: PassBy<Self>, offset: S) -> PassBy<'s, Self> {
+        let output = input + offset;
```
amcelroy, line 68: Same general comment as Aggregate block behavior.

pictorus-blocks/src/core_blocks/bit_shift_block.rs
```diff
@@ -67,65 +67,76 @@ where
...
-                            ShiftDirection::Right => input_val >> params.bits,
-                        } as $type;
-                        store.data[j][i] = res;
-                    }
-                }
-                store
+            ) -> PassBy<'s, Self> {
+                let input = input as $cast_type;
+                let output = match params.direction {
+                    ShiftDirection::Left => input << params.bits,
```
amcelroy, line 101: These can panic as well in debug or wrap in release, what about `checked_shl` or `checked_shr`? Or a panic unit test like the other blocks?

pictorus-blocks/src/core_blocks/cross_product_block.rs
```diff
@@ -66,43 +67,48 @@ pub trait Apply: Pass + Sized {
...
-        }
+/// Cross product over the flattened elements of two 3-element vectors.
+/// Both 1x3 and 3x1 matrices store their elements contiguously in the same order.
+fn cross3<S>(a: &[S], b: &[S], out: &mut [S])
+where
+    S: Scalar + Mul<Output = S> + Sub<Output = S>,
+{
+    out[0] = a[1] * b[2] - a[2] * b[1];
+    out[1] = a[2] * b[0] - a[0] * b[2];
+    out[2] = a[0] * b[1] - a[1] * b[0];
```
amcelroy, line 78: The multiplications could overflow or underflow, it doesn't look like [nalgebra accounts for this either](https://docs.rs/nalgebra/latest/src/nalgebra/base/matrix.rs.html#2074-2134). Maybe `checked_mul`? 

Nice simplification though.

pictorus-blocks/src/core_blocks/bitwise_operator_block.rs
amcelroy, line 1: This block still has promotion traits in it, are those targeted for removal in the blocks that are being updated this pass?

pictorus-blocks/src/core_blocks/bias_block.rs
```diff
@@ -166,29 +130,56 @@ mod tests {
...
-        assert_eq!(block.buffer().data, [[3.0, 4.0], [5.0, 6.0]]);
     }
 
+    test_bias_block!(f32, f64, i8, i16, i32, i64, u8, u16, u32, u64);
+
     #[test]
-    fn test_bias_matrix_to_pass() {
-        let mut block = BiasBlock::<f64, Matrix<2, 2, f64>>::default();
+    #[should_panic]
+    fn overflow_panics() {
```
amcelroy, line 179: Should we check for overflow on the negative side as well?

pictorus-blocks/src/core_blocks/gain_block.rs
```diff
@@ -130,61 +97,77 @@ impl<G: Scalar> Parameters<G> {
...
+        let output = block.process(&parameters, &context, 7);
+        assert_eq!(output, -21);
+        assert_eq!(block.buffer(), -21);
     }
 
     #[test]
-    fn test_matrix_with_to_pass() {
-        let mut block = GainBlock::<f64, Matrix<2, 2, f64>>::default();
+    #[should_panic]
+    fn overflow_panics() {
```
amcelroy, line 167: Should we check for overflow on the negative side as well?

pictorus-blocks/src/core_blocks/product_block.rs
```diff
@@ -238,4 +239,190 @@ mod tests {
...
+        let context = StubContext::default();
+        let mut block = ProductBlock::<(i32, i32), ComponentWise>::default();
+        let parameters = ParametersComponentWise::new([1.0, -1.0]);
+        let output = block.process(&parameters, &context, (4, 0));
+        assert_eq!(output, 0);
+    }
+
+    #[test]
+    #[should_panic]
+    fn int_multiply_overflow_panics() {
```
amcelroy, line 406: Should we check for overflow on the negative side as well?

pictorus-blocks/src/core_blocks/sum_block.rs
```diff
@@ -801,6 +794,120 @@ mod tests {
...
+        let parameters = Parameters {
+            operations: [SumType::Addition, SumType::Subtraction],
+        };
+        let result = block.process(&parameters, &stub_context, (3u8, 7u8));
+        assert_eq!(result, 252);
+    }
+
+    #[test]
+    #[should_panic]
+    fn addition_overflow_panics() {
```
amcelroy, line 914: Should we check for overflow on the negative side as well? Is the plan for these tests at some point to have actionable user feedback on overflows / panic behavior (Result, None, is_valid) and then these would fail and we can remove them later?

amcelroy (commented):
Generally looks good, probably some deeper questions on how we would generally handle overflows that are outside the scope of the PR. Good unit test additions, general clean up of blocks, and expansion of the types.

Will there be a codegen PR for Gain and Bias where the traits have changed?

pictorus-blocks/src/alloc_blocks/bytes_pack_block.rs
jscatena88: This was mostly to show/prove that we are truly staying in the native higher precision all the way through the block. It'd be easy enough to show that that is the case for u64 as well. What would an `f32` test be looking to show? If it was going `f32` -> `f64 -> `f32` there wouldn't be any loss of precision to show up

pictorus-blocks/src/core_blocks/abs_block.rs
jscatena88: This is just specifically for signed integers and absolute value. Since they can represent -2 ^(n-1) to 2^(n-1) - 1 the positive absolute value of  a passed in value of T::MIN can't actually be represented (e.g. -128 is i8::MIN but i8::MAX is 127). 

So a passed in value of iN::MAX will work just fine. This will definitely be the case for the other signed integer types, but since it is a a `should_panic` test they will have to be written out as individual tests per type (or macro-ized), and I figured this was enough to show the shape of the problem. 

I'll go back though and add some tests and/or comments to better describe the issue

pictorus-blocks/src/core_blocks/aggregate_block.rs
jscatena88: Yeah, I think there is definitely some design/product choices to be made for how we want to handle this. For now just using the default semantics of math operators in Rust seemed like a defensible choice when paired with these `should_panic` tests.

I really like the idea of the[ `Staturating`](https://doc.rust-lang.org/std/num/struct.Saturating.html) and [`Wrapping` ](https://doc.rust-lang.org/std/num/struct.Wrapping.html) wrapper types in the standard library, I'm not sure how difficult it would be to make them play nice with corelib though. 

But the idea of a Cast block (or maybe it'd be its own seperate type?) being able to apply those semantics is really compelling. You could have signal that is just of the type `Wrapping<Matrix<2,2,u8>>` (or maybe it would be `Matrix<2,2,Wrapping<u8>>` ?) which users could see int he debug view

pictorus-blocks/src/core_blocks/bit_shift_block.rs
jscatena88: Good call out, I'll probably bias to the same approach I was using for math operators of using the default rust operator semantics with the documenting `should_panic` tests to make it obvious what needs attention when we design a solution

pictorus-blocks/src/core_blocks/bitwise_operator_block.rs
jscatena88: I didn't set out with the express goal of removing `Promote` although I'd be happy to do so. I removed it from the Bias and Gain blocks mainly because it would have been a pain to keep it and I know it has been on the shortlist for likely removal at some point. Like I said though, I'd be very happy to rip it out everywhere though if we all agree its worth just doing now. Maybe as a follow on PR though

pictorus-blocks/src/core_blocks/cross_product_block.rs
jscatena88: As above my mental model was using default semantics with unit test documented panics until we have a designed and agreed upon approach

pictorus-blocks/src/core_blocks/sum_block.rs
jscatena88: Mainly I added them as documentation of the current behavior

pictorus-blocks/src/alloc_blocks/bytes_pack_block.rs
amcelroy: I was thinking of the lossy narrowing directions, i64 -> f32 is what an app configured for f32 signals would hit, and u64 -> f32/f64 the same way. I think I might have misunderstood the purpose of the test.

amcelroy (approved):
LGTM!

PR #71, opened 2026-08-25
Handle the case of empty bytes literal cleanly

jscatena88:
Fix for https://github.com/Pictorus-Labs/pictorus/issues/3836


amcelroy (approved):
LGTM, kind of a bummer we need to handle this case, but there isn't a newer version of the library out yet. Looks like an issue in nested in unescape_iter.

PR #72, opened 2026-08-27
Update generator blocks for rich output data types

jscatena88:
As part of data types phase 4 this PR updates the generator blocks to be able to output a richer set of types. 

The squarewave block does its actual computation in the defined scalar type, while the other blocks need to go through Float arithmetic and then cast to the expected output type.

Some parameters remain in the default float type (specifically things related to phase and timing), but anything related to the expected output values is represented in the output type (i.e. bias, amplitude, etc.)


pictorus-blocks/src/core_blocks/app_time_block.rs
```diff
@@ -11,28 +12,39 @@ impl Parameters {
 }
 
 /// Outputs the elapsed application time as a scalar value.
+///
+/// The time is read in the float type `F` and cast to the output type `T` with `as`
+/// semantics (integer outputs are truncated seconds).
```
amcelroy, line 17: Not for this PR, but I wonder if we should revisit time-keeping as part of the FFI rework. Might be a good time to swap to V2 time keeping.

amcelroy (approved):
LGTM, just an out of scope comment. Good choices for making some parameters float and other the selected type.

PR #73, opened 2026-09-09
Cast Block rounding and overflow behavior

amcelroy:
This PR gives a user the ability to fine-tune casting data types, exposing options for both rounding and overflow behavior. Both mechanisms are ZSTs passed in as generics to the cast block.

The default OverflowMode is saturate, which according to the [Rust Types doc](https://doc.rust-lang.org/rust-by-example/types/cast.html) is the behavior for float -> int. Rust `as` casting for integer -> narrow integer uses wrapping, I can change the behavior if people have strong opinions, but I suspect this is the more useful of the "what would Rust do?" behaviors. This does matter for internal callers

Changes:
- Add OverflowMode and RoundingMode traits
- Add OverflowMode impl for Panic, Wrap, and Saturate
- Add RoundingMode impl for Truncate, Nearest, Floor
- Update macro and macro invocations to account for new generics
- Update CastBlock impl with new generics
- Add unit tests to demonstrate and test behavior
- Misc. cargo fmt catches


pictorus-blocks/src/core_blocks/cast_block.rs
```diff
@@ -15,24 +16,213 @@ impl Parameters {
...
+    where
+        i128: AsPrimitive<T>,
+    {
+        if v.is_nan() {
+            return T::zero();
+        }
+        // Wrapping is defined on the integer value, so widen first. i128 covers every
+        // destination type; anything that does not fit it (the infinities, and magnitudes
+        // past 2^127) has no usable low bits left, so clamp instead.
+        match v.to_i128() {
```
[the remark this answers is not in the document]
amcelroy: The crate compiles, passes CI, and is running fine in dev mode (even on thumbv6m), not sure if this is accurate.

pictorus-blocks/src/core_blocks/product_block.rs
```diff
@@ -433,8 +433,15 @@ mod tests {
...
         let output = block.process(&p, &context, (&input, &input));
-        assert_eq!(output, &Matrix { data: [[0u8; 2]; 2] });
+        assert_eq!(
+            output,
+            &Matrix {
+                data: [[0u8; 2]; 2]
+            }
+        );
     }
 }
```
jscatena88, line 447: Not a blocker, but any clue where these formatting changes came from? Did `cargo fmt` change its behavior in an update or is there some config that is different between us developers. I just want to make sure its not the latter where we could end up with the formatting thrashing back and forth between PRs

pictorus-blocks/src/core_blocks/cast_block.rs
```diff
@@ -15,24 +16,213 @@ impl Parameters {
...
+/// - integer to integer applies the overflow mode
+/// - everything else is a plain `as` cast, where neither mode can apply.
+///
+/// The modes default to `Saturate` and `Truncate`. That reproduces `as` exactly for float
+/// to integer, but deliberately differs when narrowing between integers, where `as` wraps
+/// and `Saturate` clamps.
+///
+/// Other blocks that convert internally can still keep writing `CastElement<T>`: every one
+/// of them has a float on one side of the conversion, so the defaults and `as` agree.
+pub trait CastElement<O: Scalar, Ovf: OverflowMode = Saturate, Rnd: RoundingMode = Truncate>:
```
jscatena88, line 188: The `CastElement` trait was also used for some of the Generator blocks to be able to output a specific type (e.g. a SinewaveBlock that is set to output `i32` will still do its calculations in a float and then use `CastElement` on the way out . Since they weren't touched they'll be set to `Saturate` and `Truncate` which is probably fine for now. But we should probably at minimum document this behavior (you may have already done this in your companion PR, I'll be reading it next). And it could be worth considering if we want to expose these knobs in those generator blocks at some point.

jscatena88: Welp, probably should have read the doc comments more closely before commenting. Looks like you addressed the need for documentation since behavior is identical. Still may be worth considering adding the modes as options in the blocks though

pictorus-blocks/src/core_blocks/cast_block.rs
```diff
@@ -15,24 +16,213 @@ impl Parameters {
...
+        }
+        // Wrapping is defined on the integer value, so widen first. i128 covers every
+        // destination type; anything that does not fit it (the infinities, and magnitudes
+        // past 2^127) has no usable low bits left, so clamp instead.
+        match v.to_i128() {
+            Some(wide) => wide.as_(),
+            None => Saturate::from_float(v),
+        }
+    }
+
```
jscatena88, line 94: 100% not askling you to change this but the approach surprised me at first and isn't how I would have tried writing it at first. I wonder if if this approach is faster or slower than doing a mod operation (i.e. `(v % T::MAX) as T` ) which I think would also be mathematically correct? 

Your method may actually be more efficient though, an `i128` is pretty big on a 32 bit platform, but on the other hand once you have it I bet the `._as()` is able to just take the lower byte(s) to get the wrap value which is just memory access rather than math

pictorus-blocks/src/core_blocks/cast_block.rs
```diff
@@ -15,24 +16,213 @@ impl Parameters {
...
+}
+
+/// Round toward zero, matching `as`.
+pub struct Truncate;
+
+/// Round to the nearest whole number, with halfway cases rounding away from zero.
+pub struct Nearest;
+
+/// Round toward negative infinity.
+pub struct Floor;
```
jscatena88, line 138: Would it make sense to have a `Ceiling` mode if we have a `Floor` mode?

jscatena88 (approved):
LGTM!

pictorus-blocks/src/core_blocks/cast_block.rs
amcelroy: Yeah, that is a good idea! Let me add it in to this and the other PR.

pictorus-blocks/src/core_blocks/cast_block.rs
amcelroy: Good catch on the Generator blocks, I might add it as a follow-up PR.

pictorus-blocks/src/core_blocks/product_block.rs
amcelroy: Good question, clippy definitely changes over time. It doesn't look like we have a CI check for cargo fmt, though we install it. Let me add that as a follow-up issue.

amcelroy: https://github.com/Pictorus-Labs/pictorus-rs/issues/74

pictorus-blocks/src/core_blocks/cast_block.rs
amcelroy: I think the modulo approach is off by one, since T::MAX is 2^n - 1 (i.e. 0 to 255 is 256 values for u8). For example, we would need to modulo by 256 and not 255. The i128 conversion is annoying as it adds several instructions, but seemed like the safest way to cover i64 and u64.

PR #75, opened 2026-09-21
Rename Context trait to ModelClock

jscatena88:


pictorus-blocks/src/traits.rs
```diff
@@ -435,7 +435,7 @@ where
 /// and matrix -> matrix.
 ///
 /// Note: This currently only works for blocks that require only their parameters and inputs to
-/// determine the output. If a block requires additional context, you will need to create a bespoke
+/// determine the output. If a block requires a `ModelClock`, you will need to create a bespoke
```
amcelroy: Is this a correct replacement?

amcelroy (approved):
grep -rl "review required" . | xargs sed -i 's/review required/LGTM\!/g'

pictorus-blocks/src/traits.rs
jscatena88: Good catch!

PR #76, opened 2026-09-22
Add fsp-sys crate

Dillonmcewan:
This PR is the first part of including Renesas FSP support. It adds the `renesas-fsp-sys` crate which acts as the FFI glue layer between Rust and the Renesas FSP C API. This was largely a copy-paste of the PoC code with some of the documentation cleaned up. It also only includes the headers required for GPIO and PWM as a first pass. We can include support for other peripherals in subsequent PRs.


renesas-fsp-sys/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+homepage.workspace = true
+repository.workspace = true
+keywords.workspace = true
+categories.workspace = true
+
+# There is deliberately no `links` key, and build.rs emits no
+# `cargo:rustc-link-*`. This crate declares symbols; it never asks cargo to
+# resolve them. The FSP C sources are compiled and linked by the e2 studio
+# project that consumes the final static library, so a host `cargo build`
+# succeeds with those references left dangling.
```
amcelroy: Is this comment needed? Our other crates don't mention a links key, the README.md covers this in the Linking section.

renesas-fsp-sys/script/check-arm-abi.py
amcelroy, line 1: This script is mentioned in the README.md, but is it used anywhere?

renesas-fsp-sys/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+### Hard float
+
+A soft-float target passes `f32`/`f64` in general-purpose registers; a hard-float
+one passes them in VFP registers. The two are not interchangeable, and the linker
+does not object.
+
+Cortex-M33 RA parts have a single-precision FPU, and FSP is built
+`-mfloat-abi=hard -mfpu=fpv5-sp-d16`. The target is therefore
+`thumbv8m.main-none-eabihf` — note that `pictorus-renesas` still uses the
```
amcelroy, line 151: Whoops, guess I need to fix this, though it never crosses an FFI boundary so probably less of an issue.

renesas-fsp-sys/README.md
```diff
@@ -0,0 +1,226 @@
...
+`Tag_ABI_enum_size = small` and `Tag_ABI_VFP_args = hard` in that output are
+where the two claims above come from.
+
+## Linking
+
+Nothing here links. There is no `links` key, no `cc` invocation and no
+`cargo:rustc-link-*` — this crate emits declarations and never asks cargo to
+resolve them.
+
+So `cargo build` succeeding says nothing about whether the FSP symbols exist; it
```
amcelroy, line 161: This section is very helpful.

renesas-fsp-sys/bindings/generated.rs
amcelroy, line 1: Ah, so this is a roughly a snapshot of `bindings.rs` that is generated when `cargo build` is run, that the Rust Renesas blocks will use. As noted in the snapshot.rs and the README, this is a blind compile and always succeeds, since there is no linking step. Then the Rust blocks use`binding.rs` and generate an .a file, only then can linking be resolved when the Rust .a library compiles in the Renesas tooling.

renesas-fsp-sys/shim/bsp_api.h
amcelroy, line 1: Will this need to be updated as more peripherals are pulled in, like alternate select function enums and peripheral specific enums?

renesas-fsp-sys/script/vendor-headers.sh
```diff
@@ -0,0 +1,77 @@
...
+# Renesas FSP pack distribution. Only the interface layer (`ra/fsp/inc/api/`)
+# is vendored, and only the subset in HEADERS below -- see README.md for why
+# the instance layer cannot be included.
+#
+# Usage:
+#   script/vendor-headers.sh path/to/FSP_Packs_vX.Y.Z.zip
+#
+# The pack distribution is not redistributable and is not checked in: download
+# it from Renesas, or copy it out of an e2 studio installation under
+# `internal/projectgen/ra/packs/`. Requires `unzip`.
```
jscatena88: Is this a licensing issue? When I was playing with these I grabbed the zip file from this official Renesas github release page: https://github.com/renesas/fsp/releases/tag/v6.6.0 It could be worth adding that as an option in this comment

amcelroy (approved):
LGTM

renesas-fsp-sys/README.md
```diff
@@ -0,0 +1,226 @@
...
+script/vendor-headers.sh path/to/FSP_Packs_vX.Y.Z.zip
+```
+
+Adding a peripheral is four coordinated edits — miss one and you get either a
+stale binding or a header that fails to parse:
+
+1. add its `r_*_api.h` to `HEADERS` in `script/vendor-headers.sh`,
+2. re-run that script,
+3. add the `#include` to `wrapper.h`,
+4. refresh the snapshot and review the diff.
```
jscatena88: So step 4 is running the build command as shown in line 194 above?

I think I 90% understand what this sections is telling me to do but it could be worth spelling out the exact command sequence (and file modifcations) needed to add a peripheral or upgrade to a new FSP release

jscatena88 (approved):
LGTM! Just one or two comments on clarifying things in comments

.github/workflows/ci-tests.yaml
```diff
@@ -4,6 +4,11 @@ on:
   pull_request:
     paths:
       - "**/*.rs"
+      # renesas-fsp-sys generates its bindings from vendored C headers via a
+      # shell script, so a re-vendor or a dependency bump must run CI too.
+      - "**/*.h"
+      - "**/*.sh"
+      - "**/Cargo.toml"
```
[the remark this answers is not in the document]
Dillonmcewan: Yep, probably worth adding

renesas-fsp-sys/script/vendor-headers.sh
```diff
@@ -0,0 +1,77 @@
...
+if [[ ! -f "$packs_zip" ]]; then
+  echo "error: pack distribution not found: $packs_zip" >&2
+  exit 1
+fi
+
+# The distribution is a zip of packs; each pack is itself a zip.
+work="$(mktemp -d "${TMPDIR:-/tmp}/fsp-vendor.XXXXXX")"
+trap 'rm -rf "$work"' EXIT
+
+ra_pack_path="$(unzip -Z1 "$packs_zip" | grep -E '/Renesas\.RA\.[0-9.]+\.pack$' | head -1)"
```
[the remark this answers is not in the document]
Dillonmcewan: Punt. We can deal with this if it ever becomes an issue

renesas-fsp-sys/README.md
jscatena88: You are leaving some performance on the table but since we are compiling every artifact in your approach it remains internally consistent and isn't an issue. You just can't mix them

renesas-fsp-sys/Cargo.toml
Dillonmcewan: Agreed. Seems kind of unnecessary. I wouldn't expect there to be a `links` field so it's kind of weird to justify the absence of it

renesas-fsp-sys/script/check-arm-abi.py
Dillonmcewan: Yeah it can probably be removed

renesas-fsp-sys/README.md
Dillonmcewan: I suspect we will probably deprecate the old Renesas approach, so we can deal with it at that point

renesas-fsp-sys/shim/bsp_api.h
Dillonmcewan: Correct

renesas-fsp-sys/README.md
Dillonmcewan: Yep. Probably worth adding a more complete example/instructions

renesas-fsp-sys/script/vendor-headers.sh
Dillonmcewan: Agreed, this comment is a little confusing. I don't think there's really a licensing issue, more just that we don't want to check in a giant zip file here. Worth at least adding the link

